PR #7, opened 2018-07-30
Implement calculation of suggestions for any text position

boq:
Makes calculation of suggestions in middle of command less expensive.

Also changes nodes in context from `Map` to `List` - there were never used as map, so extra overhead was not needed.


boq:
@Dinnerbone done

PR #10, opened 2018-09-26
Long arguments

Earthcomputer:
Adds a long argument type to be along with the other default primitive argument types (integer, double, etc.). I have needed this a couple of times already in a couple of Minecraft mods and thought this ought to go in brigadier itself rather than create it myself for every mod I make.

### DIFF
--- a/src/main/java/com/mojang/brigadier/arguments/LongArgumentType.java
+++ b/src/main/java/com/mojang/brigadier/arguments/LongArgumentType.java
@@ -1,0 +1,87 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+package com.mojang.brigadier.arguments;
+
+import com.mojang.brigadier.StringReader;
+import com.mojang.brigadier.context.CommandContext;
+import com.mojang.brigadier.exceptions.CommandSyntaxException;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+public class LongArgumentType implements ArgumentType<Long> {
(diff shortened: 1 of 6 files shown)


src/main/java/com/mojang/brigadier/arguments/LongArgumentType.java
```diff
@@ -0,0 +1,87 @@
...
+
+import com.mojang.brigadier.StringReader;
+import com.mojang.brigadier.context.CommandContext;
+import com.mojang.brigadier.exceptions.CommandSyntaxException;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+public class LongArgumentType implements ArgumentType<Long> {
+    private static final Collection<String> EXAMPLES = Arrays.asList("0", "123", "-123");
```
Pokechu22: You don't declare `getExamples`, so this ends up being unused.

Earthcomputer: Same in `IntegerArgumentType` also, I just copied from that and didn't notice. Both need to be fixed.

boq: `IntegerArgumentType` was already fixed, though after you commit was merged: https://github.com/Mojang/brigadier/blob/master/src/main/java/com/mojang/brigadier/arguments/IntegerArgumentType.java#L90

Earthcomputer: See #16

PR #14, opened 2018-09-27
Add some quality of life improvements to `StringReader`

Yurihaia:
Added 2 new useful functions to `StringReader`

1. overload of `StringReader::expect` that accepts a string, and expects the string
2. `StringReader::expectOption` which expects  one of a list of String

### DIFF
--- a/src/main/java/com/mojang/brigadier/StringReader.java
+++ b/src/main/java/com/mojang/brigadier/StringReader.java
@@ -239,4 +239,21 @@ public class StringReader implements ImmutableStringReader {
         }
         skip();
     }
+
+    public void expect(final String s) throws CommandSyntaxException {
+        if(!canRead(s.length()) || !getRemaining().startsWith(s)) {
+            throw CommandSyntaxException.BUILT_IN_EXCEPTIONS.readerExpectedSymbol().createWithContext(this, s);
+        }
+        this.cursor += s.length();
+    }
+
+    public String expectOption(final String ...opts) throws CommandSyntaxException {
+        for(String s: opts) {
+            try {
+                expect(s);
+                return s;
+            } catch (CommandSyntaxException e) {}
+        }
+        throw CommandSyntaxException.BUILT_IN_EXCEPTIONS.readerExpectedSymbols().createWithContext(this, opts);
+    }
 }
--- a/src/main/java/com/mojang/brigadier/StringReader.java
+++ b/src/main/java/com/mojang/brigadier/StringReader.java
@@ -241,19 +241,28 @@ public class StringReader implements ImmutableStringReader {
     }
 
     public void expect(final String s) throws CommandSyntaxException {
-        if(!canRead(s.length()) || !getRemaining().startsWith(s)) {
+        if (!canRead(s.length()) || !getRemaining().startsWith(s)) {
             throw CommandSyntaxException.BUILT_IN_EXCEPTIONS.readerExpectedSymbol().createWithContext(this, s);
         }
         this.cursor += s.length();
     }
 
-    public String expectOption(final String ...opts) throws CommandSyntaxException {
-        for(String s: opts) {
+    public String expectOption(final String... opts) throws CommandSyntaxException {
+        String out = "";
+        final int start = this.getCursor();
+        for (String s: opts) {
             try {
+                this.setCursor(start);
                 expect(s);
-                return s;
-            } catch (CommandSyntaxException e) {}
+                if (out.length() < s.length()) {
+                    out = s;
+                }
+            } catch (final CommandSyntaxException ignored) {}
         }
-        throw CommandSyntaxException.BUILT_IN_EXCEPTIONS.readerExpectedSymbols().createWithContext(this, opts);
+        if (out.length() == 0) {
+            throw CommandSyntaxException.BUILT_IN_EXCEPTIONS.readerExpectedSymbols().createWithContext(this, opts);
+        }
+        this.setCursor(start + out.length());
+        return out;
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/com/mojang/brigadier/StringReader.java
```diff
@@ -239,4 +239,21 @@ public void expect(final char c) throws CommandSyntaxException {
...
     }
+
+    public void expect(final String s) throws CommandSyntaxException {
+        if(!canRead(s.length()) || !getRemaining().startsWith(s)) {
+            throw CommandSyntaxException.BUILT_IN_EXCEPTIONS.readerExpectedSymbol().createWithContext(this, s);
+        }
+        this.cursor += s.length();
+    }
+
+    public String expectOption(final String ...opts) throws CommandSyntaxException {
```
Dinnerbone: I'm a little unsure about this. If you did `expectOption("foo", "bar", "footprint", "barstool")` on the string `"footprint"` you'd get `"foo"` as a return. Technically correct, but would that be clear? It's also not nearly as efficient as just taking the string and comparing afterwards.

Can you provide me some use cases of how this is intended to be used? I can only imagine trying to parse literals, but that's handled in another way.

Dinnerbone: Style issue: do `final String... opts`

(I'm sorry - this is my fault for not commiting a style guide yet. I'll do that as soon as I can.)

src/main/java/com/mojang/brigadier/StringReader.java
```diff
@@ -239,4 +239,21 @@ public void expect(final char c) throws CommandSyntaxException {
...
+        for(String s: opts) {
```
Dinnerbone: Style issue: do `for (String s: opts) {`

(I'm sorry - this is my fault for not commiting a style guide yet. I'll do that as soon as I can.)

src/main/java/com/mojang/brigadier/StringReader.java
```diff
@@ -239,4 +239,21 @@ public void expect(final char c) throws CommandSyntaxException {
...
+            try {
+                expect(s);
+                return s;
+            } catch (CommandSyntaxException e) {}
```
Dinnerbone: Style issue: Name the exception `ignored` to make it clear we don't care for it, and make it final.

(I'm sorry - this is my fault for not commiting a style guide yet. I'll do that as soon as I can.)

src/main/java/com/mojang/brigadier/StringReader.java
MrYurihi: OK. I guess it makes sense for expectOption to match the longest instead of the first

MrYurihi: It would be used for things like matching `}` or `,`.

src/main/java/com/mojang/brigadier/StringReader.java
```diff
@@ -239,4 +239,30 @@ public void expect(final char c) throws CommandSyntaxException {
...
+        for (String s: opts) {
+            try {
+                this.setCursor(start);
+                expect(s);
+                if (out.length() < s.length()) {
+                    out = s;
+                }
+            } catch (final CommandSyntaxException ignored) {}
+        }
+        if (out.length() == 0) {
```
Dinnerbone: If one of the `opts` is `""`, this will inadvertantly throw an error. Please add a unit test to cover this - and either throw a deliberate "you can't expect nothing" error (IllegalArgumentException?) or handle this properly (default to null, test for null for your match)

MrYurihi:
There we go. I made all of the requested changes (I hope)

src/main/java/com/mojang/brigadier/StringReader.java
Sollace: > and make it final.

Now that is a style choice I've never heard of before...?

MrYurihi: I suppose the JVM handles final things better, so I guess it is a good style choice?

src/main/java/com/mojang/brigadier/StringReader.java
```diff
@@ -239,4 +239,30 @@ public void expect(final char c) throws CommandSyntaxException {
...
+            try {
+                this.setCursor(start);
+                expect(s);
+                if (out == null || out.length() < s.length()) {
+                    out = s;
+                }
+            } catch (final CommandSyntaxException ignored) {}
+        }
+        if (out == null) {
+            throw CommandSyntaxException.BUILT_IN_EXCEPTIONS.readerExpectedSymbols().createWithContext(this, opts);
```
Marcono1234: Might have to reset the cursor when an exception is thrown, see #33

Marcono1234:
Maybe it would be useful to have a `expectOption` overload which accepts some kind of Collection, for example List. Then the current implementation could call it with `expectOption(Arrays.asList(opts))` which is not that expensive since the created list is backed by the array.

Sollace:
@Marcono1234 I'm not fond of any options that require you to wrap an array in a collection.

That said, I could see an overload that takes a collection being useful in some situations. My approach would be to make it as generalised as possible, though, so you can take any object you want, not just lists.

```
String expectOption(final Collection<String> opts);

String expectOption(final String... opts);
```

```
        expectOption("1", "2", "3");
        expectOption(new ArrayList<>());
        expectOption(new HashSet<>());
```

MrYurihi:
@Dinnerbone Is there anything else I need to change for this to go through?

peterix:

peterix:
OOF. I guess we cannot reopen. And we can't rerun checks without doing odd things.

Sorry :(

PR #15, opened 2018-09-27
added support for term-type string arguments with fixed set of choices

gnembon:
Simple way to support a fixed set of choices and let the parser fail when requirements are not met.
Would replace calling literal multiple times or the need of using customized argument type class.


boq:
Would be much more useful as multiple literal replacement if it implemented `listSuggestions` and `getExamples` based on actual options (examples are used for detecting ambiguities, [see here](https://github.com/Mojang/brigadier/blob/master/src/main/java/com/mojang/brigadier/tree/CommandNode.java#L106)). I will probably PR some utilities for simplifying suggestion handling tomorrow.

Also, please consider splitting that to separate class, if amount of special casing increases (that new optional field already seems bad, IMO)

gnembon:
I added getExamples as well.
I am leaving the decision of splitting it into its own argument type class to the owners of the repo. As I said, from the usecase perspective, I don't see much difference semantically between word() and term() thus feel appropriate to have them together.

boq:
About examples: let's say you (incorrectly) define four children: literals `a`, `b`, `b`, `c`. In such case, ambiguity will be detected, since literals return [themselves as example](https://github.com/Mojang/brigadier/blob/master/src/main/java/com/mojang/brigadier/tree/LiteralCommandNode.java#L122) - one of the `b` nodes will accept other one's example.

Now split that into two term strings: `a`+`b` and `b`+`c` - in this case ambiguity will remain undetected, since argument always returns `["foo bar baz"]` as example (not even `["foo","bar","baz"]`!) and other node will not accept it (`"foo bar baz" != "a" & "foo bar baz" != "b"`).

As for separate class: it already starts to [special case single enum value](https://github.com/Mojang/brigadier/pull/15/files#diff-f5a5a69e9d59ce7f1778738ea00441b5R92), so it conflicts more and more with open/closed principle (which is not required, but is quite nice guideline). Also, other options describe parsing strategy, while new one adds validation (one could argue that `QUOTABLE_PHRASE` could also use limited option set).

boq:
I guess second class is not needed, if limitation of value is applied in all cases - it would be somehow similar to `IntegerArgumentType` behaviour, which also has optional bounds.

kashike:
I think it might be good to have this as a separate argument type/class - `StringChoiceArgumentType` or something.

Pokechu22:
I also would go for a separate class, and possibly utility methods to convert from/to an enum (but I'm not 100% sure what those would look like).

> EDIT: Of course, there is extra consideration that in Minecraft code we have to serialize arguments, which complicates things little bit, since sending strings over network (multiple times, in node is reused) is not really nice. That would suggest "baking" limitations into class instead of instance (via abstract method?), since it would me more efficient to just define separate argument type based on this one.

I think it would be more useful to have a general way of serializing a list over the network, just so that there's a type that can be used on the server without the client having to know about the custom type.  Of course for more frequent uses it would be better to have specific subclasses.

Pokechu22:
To give an example of what I said earlier, consider `/playsound` and its categories.  Currently, each category is its own literal, and that means that the the arguments after that are repeated each time.  A single argument for that would make the code much cleaner, and also reduce duplication.

boq:
It sounds like wrapper around `Map<String, T>` (presumably with map that is accessed during runtime, since in MC sounds can be reloaded and commands are not). May be useful to have that as an abstract base class. Anyway, it's probably separate concern from value-limited string.

Pokechu22:
I'm referring to the category/source (which is a fixed set of values, `master`/`music`/`record`/`weather`/`block`/`hostile`/`neutral`/`player`/`ambient`/`voice`, and is backed by an `enum`), not the list of sounds themselves (e.g. `minecraft:ambient.cave`).

src/main/java/com/mojang/brigadier/arguments/StringArgumentType.java
```diff
@@ -57,7 +84,16 @@ public String toString() {
...
-        return type.getExamples();
+        return type == StringType.TERM ? options : type.getExamples();
+    }
+
+    @Override
+    public <S> CompletableFuture<Suggestions> listSuggestions(final CommandContext<S> context, final SuggestionsBuilder builder) {
+        if (type != StringType.TERM) {
+            return Suggestions.empty();
+        }
+        options.stream().filter((s) -> s.startsWith(builder.getRemaining().toLowerCase())).sorted().forEachOrdered(builder::suggest);
```
Marcono1234: This assumes that the `options` values are all lowercase, but this is not guaranteed, is it?

gnembon: I agree. I will change it so that it matches behaviour of some other examples that Mojang uses (selection is case sensitive, suggestions - insensitive)

gnembon:
Ok then. Lemme rewrite it as a subclass of StringArgumentType. Should then the 'getter' just be using getString from the super? or should it have its own (getTerm or whatever)? I think as a subclass, using getString would be sufficient and appropriate?
RFC guys.

Marcono1234:
Should `toLowerCase` use the default locale (which is currently the case) or the `ROOT` locale? The used locale affects how certain characters are converted, see [https://unicode.org/Public/UNIDATA/SpecialCasing.txt](https://unicode.org/Public/UNIDATA/SpecialCasing.txt). I assume `ROOT` would be better since it does not depend on the system locale this way?

gnembon:
Closing - seems like if there is support to that idea, it would be more so as a subclass of StringArgumentType for which I made a pull req as well.

PR #18, opened 2018-09-28
Add orDefault methods to help with optional arguments

Earthcomputer:
Right now there is a problem with lots of consecutive optional arguments; the cleanest way to do it currently is to create a very repetitive list of commands like [this](https://github.com/Earthcomputer/Quick-1.13.1-Carpet/blob/master/src/main/java/quickcarpet/StructureCommand.java). What would be better is to define a `Command<S>` once and pass it to multiple different `.execute()` clauses. Then, each optional argument can have a default value.
Let me give an example of optional arguments with this PR:
```java
Command<S> foo = context -> doFoo(getInteger(context, "bar"), getIntegerOrDefault(context, "baz", 0));
dispatcher.register(literal("foo")
  .then(argument("bar", integer())
    .executes(foo)
    .then(argument("baz")
      .executes(foo))));
```
This has the advantage of removing a lot of code duplication, especially for long chains of optional arguments like the previously linked `/structure` command.

Just as a clarification, it technically *was* possible to emulate the orDefault behaviour by surrounding the context.getArgument() method with a try-catch block, but this was quite verbose and a series of getOrDefault methods is much cleaner in my opinion.

### DIFF
--- a/src/main/java/com/mojang/brigadier/arguments/BoolArgumentType.java
+++ b/src/main/java/com/mojang/brigadier/arguments/BoolArgumentType.java
@@ -27,6 +27,10 @@ public class BoolArgumentType implements ArgumentType<Boolean> {
         return context.getArgument(name, Boolean.class);
     }
 
+    public static boolean getBoolOrDefault(final CommandContext<?> context, final String name, final boolean defaultValue) {
+        return context.getArgumentOrDefault(name, defaultValue, Boolean.class);
+    }
+
     @Override
     public Boolean parse(final StringReader reader) throws CommandSyntaxException {
         return reader.readBoolean();
--- a/src/main/java/com/mojang/brigadier/arguments/DoubleArgumentType.java
+++ b/src/main/java/com/mojang/brigadier/arguments/DoubleArgumentType.java
@@ -37,6 +37,10 @@ public class DoubleArgumentType implements ArgumentType<Double> {
         return context.getArgument(name, Double.class);
     }
 
+    public static double getDoubleOrDefault(final CommandContext<?> context, final String name, final double defaultValue) {
+        return context.getArgumentOrDefault(name, defaultValue, Double.class);
+    }
+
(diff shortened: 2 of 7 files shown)


JarvisCraft:
Why are you using primitives in some cases but their wrappers in others?

Earthcomputer:
I am copying what was there already in the non-defaulted method of each class

Earthcomputer:
This might be okay for a small number of optional arguments, but gets prohibitively verbose when there are six or seven in a row. The other solution is to create overloads of the `doCommand` method, but that's even more verbose.

Marcono1234:
Maybe I am not understanding the proposed changes correctly, but it looks like what Earthcomputer is proposing does not require defining a general default value. The default value is only defined in the command.
Therefore custom types (I assume you mean `ArgumentType` subclasses) only have to implement the <code>get<i>TypeName</i></code> and <code>get<i>TypeName</i>OrDefault</code> methods.

@Earthcomputer, what do you think about naming it <code>getOptional<i>TypeName</i></code> and then returning an `Optional`? This allows, besides using a default value, also a default value supplier, which might be useful if the value creation is expensive. For example:

    static Optional<String> getOptionalString(final CommandContext<?> context, final String name) {
        ...
    }
<b></b>

    Command<S> foo = context -> doFoo(getOptionalString(context, "baz").orElseGet(() -> createStringExpensive()));

Earthcomputer:
I personally don't like `Optional`, the syntax is clunky and if you're worried about NPEs, `@Nullable` is a nice alternative.

Marcono1234:
In this case I was not worried about NPEs, but wanted to make it easier to get a default value (or create a custom exception). Otherwise we might need additionally to your `...OrDefault(CommandContext<?>, String, T)` a `...OrDefault(CommandContext<?>, String, Supplier<T>)` method (with `T` being for example `boolean`).

PR #24, opened 2018-10-02
Formatted license

JarvisCraft:
This commit simply makes some changes on `LICENSE` formatting by *adding more lines to separate sentences* and *adding quotes to **AS IS** termina* as in default GitHub's MIT license template.


IkeKap:
Did you set yourself a specific character limit per line? Or did you mostly eyeball it?
Looks good and a lot cleaner (especially on a device with a smaller screen like mobile)

JarvisCraft:
@IkeKap, I've used GitHub's default MIT license template

Dinnerbone:
Other than linebreaks this actually does change the LICENSE in 2 places:

> (the Software)
> (the "Software")

> \*AS IS\*
> "AS IS"

I'm fine with that, so we'll accept it! But writing it down here to save anyone else some time :)

PR #27, opened 2018-10-07
Fix very minor grammer mistakes

yourWaifu:
These are very minor. I'm not 100% sure if they are valid but I did a check in Grammarly before making this commit. There are still some that Grammarly thinks is wrong but I didn't really agree with it. Other then that, you may close this pull request since I'm not even sure if you guys will even accept it since for most people grammar mistakes aren't high on the priority list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,13 +50,13 @@ contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additio
 # Usage
 At the heart of Brigadier, you need a `CommandDispatcher<S>`, where `<S>` is any custom object you choose to identify a "command source".
 
-A command dispatcher holds a "command tree", which are a series of `CommandNode` which represent the various possible syntax that form a valid command.
+A command dispatcher holds a "command tree", which is a series of `CommandNode`s that represent the various possible syntax options that form a valid command.
 
 ## Registering a new command
 Before we can start parsing and dispatching commands, we need to build up our command tree. Every registration is an append operation,
 so you can freely extend existing commands in a project without needing access to the source code that created them.
 
-Command registration also encourages use of a builder pattern to keep code cruft to a minimum.
+Command registration also encourages the use of a builder pattern to keep code cruft to a minimum.
 
 A "command" is a fairly loose term, but typically it means an exit point of the command tree.
 Every node can have an `executes` function attached to it, which signifies that if the input stops here then this function will be called with the context so far.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,13 +50,13 @@ contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additio
 # Usage
 At the heart of Brigadier, you need a `CommandDispatcher<S>`, where `<S>` is any custom object you choose to identify a "command source".
 
-A command dispatcher holds a "command tree", which are a series of `CommandNode` which represent the various possible syntax that form a valid command.
+A command dispatcher holds a "command tree", which are a series of `CommandNode` which represents the various possible syntax that forms a valid command.
```
Pokechu22: I'm not sure I agree with this.  The issue here is that `CommandNode` isn't written as plural, but it would be for "series of"; you say "a series of objects which represent something" and not "a series of object which represents something".

yourWaifu: Ok, I'm not an expert in this, as I'm not an English teacher but a programmer, and English is actually my 2nd language, Chinese being my 1st language, so take this with a grain of salt.

But looking at the sentence, I think maybe turning

> A command dispatcher holds a "command tree", which are a series of `CommandNode` **which represent** the various possible syntax that form a valid command.

to

> A command dispatcher holds a "command tree", which are a series of `CommandNode` **representing** the various possible syntax that form a valid command.

might work better in this case.

yourWaifu: Ok, I spent some time with others thinking about this, and we ended up with this. 

> A command dispatcher holds a "command tree", which is a series of `CommandNode`s that represent the various possible syntax options that form a valid command.

Is this Good?

liach:
Don't trust grammarly too much. Some acceptable styles, e.g. passive voice, are identified by Grammarly as issues.

README.md
liach: uh, you should have consulted an SAT writing (grammar) teacher before fixing grammar then.

README.md
```diff
@@ -100,7 +100,7 @@ When a command is actually run, it can access these arguments in the context pro
 ## Parsing user input
 So, we've registered some commands and now we're ready to take in user input. If you're in a rush, you can just call `dispatcher.execute("foo 123", source)` and call it a day.
 
-The result of `execute` is an integer returned by the command it evaluated. Its meaning varies depending on command, and typically will not be useful to programmers.
+The result of `execute` is an integer returned by the command it evaluated. Its meaning varies depending on the command, and typically will not be useful to programmers.
```
liach: should be `a commands it has evaluated`

boomboompower: It should be the not a

liach: Nope. Here is an example:
```
I ate an apple. The apple was sweet.
```
Similarly, these sentences below are also valid:
```
The method have evaluated a command. The command defines the meaning of the integer.
```
@boomboompower Thus, I disagree with your proposal.

Pokechu22: "The method have evaluated a command." does not sound valid to me.

liach: Oh, it was just a minor issue I've accidentally introduced.

liach: Fixed. @Pokechu22

yourWaifu: I think I agree with @boomboompower, it doesn't sound right to me without the "the". However, lets forget about that since I think this will work.

> The result of `execute` is an integer that was returned from an evaluated command. The meaning of this integer depends on the command, and will typically not be useful to programmers.

JarvisCraft:
Oh lol, I've discovered right now that PR is named "grammEr"

PR #31, opened 2018-10-09
Allow scientific notation of numbers and `+` sign

NeunEinser:
This is a simple fix to allow for scientific notation of floating point numbers.
This commit would also fix the following ticket on the Minecraft Bug tracker: https://bugs.mojang.com/browse/MC-130925

### DIFF
--- a/src/test/java/com/mojang/brigadier/StringReaderTest.java
+++ b/src/test/java/com/mojang/brigadier/StringReaderTest.java
@@ -388,33 +388,33 @@ public class StringReaderTest {
     
     @Test
     public void readDouble_explicitPositive() throws Exception {
-        final StringReader reader = new StringReader("+1234567890");
-        assertThat(reader.readDouble(), is(+1234567890.0));
-        assertThat(reader.getRead(), equalTo("+1234567890"));
+        final StringReader reader = new StringReader("+123");
+        assertThat(reader.readDouble(), is(+123.0));
+        assertThat(reader.getRead(), equalTo("+123"));
         assertThat(reader.getRemaining(), equalTo(""));
     }
     
     @Test
     public void readDouble_exponent() throws Exception {
-        final StringReader reader = new StringReader("1.23E4");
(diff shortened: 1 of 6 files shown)


Earthcomputer:
:+1: for splitting `isAllowedNumber` and allowing the + sign in both

NeunEinser:
> Tests? :)

Good point. Added tests for double and float with both positive and negative exponent. That should be sufficient, I hope :)

NeunEinser:
I now added a method that is specific for integer numbers. The old method still exists and does the same thing as before (so backwards compatibility is supported). Both methods do accept the + sign now.

NeunEinser:
Did that. Makes sense to provide this information to the user.
I'm a bit unsure about the recommended max line length of this project but I found it to be more readable if I split the examples of double and float into two lines now, that the list has become longer.

src/test/java/com/mojang/brigadier/StringReaderTest.java
```diff
@@ -421,14 +469,46 @@ public void readFloat_withDecimal() throws Exception {
         assertThat(reader.getRead(), equalTo("12.34"));
         assertThat(reader.getRemaining(), equalTo(""));
     }
-
+
```
liach: Extra spaces

NeunEinser:
Thank you! Unfortunately, Mojang doesn't seem to care about user contributions anymore and this is effectively a dead repo.

liach:
i think they will probably take it for 1.14's additional releases or more likely, 1.15 snapshots

NeunEinser:
Well, this pr is from well before the 1.14 snapshot phase, so I am not that hopeful. In fact, the latest Mojang activity on user prs or issues was on Oct 14, 2018.
The only pr that was worked on since then was an internal one, probably because boq told Dinnerbone to look at it when they were at work.

peterix:
Closing and reopening to rerun checks. Hopefully this works... 🤞

PR #34, opened 2018-10-10
term type argument as a subclass from string type argument

gnembon:
Another option for term-like string argument, as a subclass of StringArgumentType. Reasoning - code is much cleaner, yet still one would retrieve its value via getString as it is technically nothing but a string from a predefined collection, and I don't think making it a totally separate class from StringArgumentType makes sense as the returned value is a string and that object type you would expect. The fact that it is a subclass makes it clear in my opinion of a getter method as getString.

Please send feedback regarding this implementation.

In case the choice of the word 'term' to describe it is not clear, I shall direct to the following definition
https://en.wikipedia.org/wiki/Terminology

SYNOPSIS (and why this type is very useful):
```        
        dispatcher.register(literal("test").
                then(argument("first", term("foo","bar")).
                        executes( (c)-> test(c, getString(c, "first")+" 1"))).
                then(argument("second", term("baz","quux")).
                        executes( (c)-> test(c, getString(c, "second")+" 2"))));
```


src/main/java/com/mojang/brigadier/exceptions/BuiltInExceptionProvider.java
```diff
@@ -28,6 +28,8 @@
 
     DynamicCommandExceptionType readerInvalidEscape();
 
+    DynamicCommandExceptionType readerInvalidTerm();
```
Marcono1234: Might be better to not prefix it with `reader` since it is not the reader which parses terms (it only parses strings).

src/main/java/com/mojang/brigadier/arguments/StringArgumentType.java
```diff
@@ -13,7 +13,7 @@
 public class StringArgumentType implements ArgumentType<String> {
     private final StringType type;
 
-    private StringArgumentType(final StringType type) {
+    StringArgumentType(final StringType type) {
```
Marcono1234: Would not matter much, but what do you think about using `protected` as visibility instead? This would allow subclasses outside of this package (which should possibly be the case for the other argument types as well, but that is a different topic).

gnembon: agreed, fixed

src/main/java/com/mojang/brigadier/exceptions/BuiltInExceptionProvider.java
gnembon: agreed. fixed.

src/main/java/com/mojang/brigadier/arguments/TermArgumentType.java
```diff
@@ -0,0 +1,59 @@
...
+        final String term = reader.readUnquotedString();
+        if (!options.contains(term)) {
+            throw CommandSyntaxException.BUILT_IN_EXCEPTIONS.termInvalid().createWithContext(reader, term);
+        }
+        return term;
+    }
+
+    @Override
+    public String toString() {
+        return "term("+options+")";
```
kashike: `"term(" + options + ")"`

gnembon: I assumed you meant extra spcaces. Fixed :)

src/main/java/com/mojang/brigadier/arguments/StringArgumentType.java
```diff
@@ -87,7 +87,8 @@ private static String escape(final String input) {
     public enum StringType {
         SINGLE_WORD("word", "words_with_underscores"),
         QUOTABLE_PHRASE("\"quoted phrase\"", "word", "\"\""),
-        GREEDY_PHRASE("word", "words with spaces", "\"and symbols\""),;
+        GREEDY_PHRASE("word", "words with spaces", "\"and symbols\""),
+        TERM("predefined_tokens","red","green","blue"),;
```
kashike: `predefined_token` instead of `predefined_tokens`

gnembon: applied.

peterix:

PR #36, opened 2018-10-13
Update

thebeast-p:


MiniDigger:
If you want to translate the readme, you should do so in a separate file.
But I highly doubt that this is needed.

sjx233:
So you copied the beginning of `README.md` to a translator with the target language set to Hindi, copied the result back overwriting existing content, clicked on [Propose file change] and accidentally created this PR?

PolarizedIons:
@sjx233 not just that, but a new account, doing the same to a bunch of repos in a short time frame. I have doubts that this was even manually done :man_shrugging:

PR #37, opened 2018-10-19
Cleaning up according to infered code style, and fixed minor typos

NeunEinser:
Some minor cleanups IntelliJ reported to me with the infered style settings when working on a different branch.
Please let me know if I interpreted some style choices wrong or something should be done differently. Overall it should be more consistent with the rest of the project.


src/test/java/com/mojang/brigadier/tree/LiteralCommandNodeTest.java
```diff
@@ -95,7 +95,7 @@ public void testSuggestions() throws Exception {
         assertThat(food.isEmpty(), is(true));
 
         final Suggestions b = node.listSuggestions(contextBuilder.build("b"), new SuggestionsBuilder("b", 0)).join();
-        assertThat(food.isEmpty(), is(true));
```
NeunEinser: Found that because `b` was an unsed variable before. Copy/paste mistake here.

Issue #32, opened 2018-10-09
[Bug] Merging of sibling nodes with same name is faulty


Marcono1234:
It appears it is intended that sibling nodes with the same name exist (see this [comment](https://github.com/Mojang/brigadier/issues/29#issuecomment-427676491)). However the way they are treated is in my opinion rather unintuitive.

The method [`CommandNode.addChild(CommandNode<S>)`](https://github.com/Mojang/brigadier/blob/7ee589b29b7c72c423c1549f1edcb5c89981291a/src/main/java/com/mojang/brigadier/tree/CommandNode.java#L77) replaces the command of the existing node and adds the children of the new node to the existing one. This is, at least from my point of view, not the wanted behavior since 
- The old command is lost
- Other attributes of the new node like the requirement are lost

It might be best to either remove this merging and allow siblings with the same name, or to prevent them when building.

# Example code

    CommandDispatcher<Object> dispatcher = new CommandDispatcher<>();
        
    dispatcher.register(
        literal("foo")
        .then(
            argument("bar", word())
            .executes(c -> {
                System.out.println("string");
                return 1;
            })
        )
        .then(
            argument("bar", bool())
            .executes(c -> {
                System.out.println("bool");
                return 1;
            })
        )
    );
    
    dispatcher.execute("foo not_a_bool", null);

Dinnerbone:
Names must be unique, so we can't allow multiple with the same name. It's used in a variety of places as a key, such as pathing in the tree, serialization, or simply getting values out of the context.

Merging is supported as a design principle (which I really must write down somewhere!). It must be supported to "enhance" a command after it has already been registered, to allow mods to easily extend commands in Minecraft (and I'm sure other applications/games will find that useful too).

That said, merging must be made smarter - I totally agree with this. We should probably do some sanity checking when a merge happens, and throw an exception if it's nonsense - if anything other than the children changed. For example in your code, the second `.then(argument("bar"), bool())` should throw an exception as the very type of the argument has changed, and that should not be allowed.

Marcono1234:
What do you think about having separate methods for this modification instead? For example `ArgumentCommand` could have the methods

* `addArgument(ArgumentType<T>)`
    Adds the given argument type, which has to have the same type as the argument node. Argument types would then be stored in a list and the respective methods would go through the other arguments in this list if one fails.
* `setArguments(List<ArgumentType<T>>)`
    Replaces all currently set argument types.

[...]

PR #47, opened 2018-11-11
Add some javadoc comments to give a rough outline

I-Al-Istannen:
Sorry for the large PR, but I wasn't sure if and where to split it up.

## Content
This PR should only contain javadoc, but that for about any public member I could find.
The goal is to hopefully make navigating the API outside of CommandDispatcher a little less annoying and clarify the usage. What is still missing are some proper walkthroughs or tutorials that actually walk you through achieving some goal, but IMHO useful javadoc is a good first step.

I *hope* the doc I wrote is useful but contributions, questions and clarifications are more than welcome!

## Problems
The docs still have around 6 places marked with `TODO: XXX`, indicating that I had no idea why a certain choice was made and therefore could not write docs for it. The other place is in the Suggestions create/merge methods, as I know how they work but not what they are used for. Here is a list with all TODOs:
- [ ] `Command#run` throws a Command**Syntax**Exception, but is *executed* and not parsed here
- [ ] `CommandContextBuilder` contains a `CommandDispatcher` which is never used and not present in the `CommandContext`
- [ ] Why does the `CommandContext(Builder)` contain a *list* of nodes, when the `withNode` method replaces all relevant fields but only keeps the old node around? And the only use case for `getNodes` I have found is to get the node for the attached command, so the last one. The only references in the project also just fetch *one* node.
- [ ] `StringRange` constructor: Why allow ranges where the start is greater than the end? Something like `start = 20, end = 5` might not be ideal.
- [ ] `CommandSyntaxException` seems to be more than just a *syntax* exception. Why is only syntax in the name then?

- [ ] Some clarification about how `Suggestions#merge` and `Suggestions#create` work/are used for

## Other
Clarifications and change requests/comments are welcome!

### DIFF
--- a/src/main/java/com/mojang/brigadier/Command.java
+++ b/src/main/java/com/mojang/brigadier/Command.java
@@ -6,9 +6,30 @@ package com.mojang.brigadier;
 import com.mojang.brigadier.context.CommandContext;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 
+/**
+ * The functional interface actually representing a command to execute.
+ * <p>
+ * This interface represents the code that will be run, when the command is executed.
+ *
+ * @param <S> the type of the command source this command will be called with
+ */
 @FunctionalInterface
 public interface Command<S> {
+    /**
+     * A static constant you can use when the command completed successfully.
+     * <p>
+     * As the return value means what you want it to mean, this is merely a suggestion to make the code a bit more
+     * expressive, if you follow it.
+     */
     int SINGLE_SUCCESS = 1;
 
+    /**
+     * Executes the command using the given {@link CommandContext} to supply information to the command.
+     *
+     * @param context the command context to use for supplying all information the command will need
+     * @return the result of executing the command. The value is arbitrary, though {@link #SINGLE_SUCCESS} is a
+     * suggestion
+     * @throws CommandSyntaxException TODO: Why does it throw a <strong>Syntax</strong> exception here?
+     */
     int run(CommandContext<S> context) throws CommandSyntaxException;
 }
--- a/src/main/java/com/mojang/brigadier/CommandDispatcher.java
+++ b/src/main/java/com/mojang/brigadier/CommandDispatcher.java
@@ -6,6 +6,7 @@ package com.mojang.brigadier;
 import com.mojang.brigadier.builder.LiteralArgumentBuilder;
 import com.mojang.brigadier.context.CommandContext;
 import com.mojang.brigadier.context.CommandContextBuilder;
+import com.mojang.brigadier.context.StringRange;
 import com.mojang.brigadier.context.SuggestionContext;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import com.mojang.brigadier.suggestion.Suggestions;
@@ -435,8 +436,8 @@ public class CommandDispatcher<S> {
      * listed as multiple entries: the parent node, and the child nodes.
      * For example, a required literal "foo" followed by an optional param "int" will be two nodes:</p>
      * <ul>
-     *     <li>{@code foo}</li>
-     *     <li>{@code foo <int>}</li>
+     * <li>{@code foo}</li>
+     * <li>{@code foo <int>}</li>
      * </ul>
      *
      * <p>The path to the specified node will <b>not</b> be prepended to the output, as there can theoretically be many
@@ -481,7 +482,7 @@ public class CommandDispatcher<S> {
      * These forms may be mixed and matched to provide as much information about the child nodes as it can, without being too verbose.
      * For example, a required literal "foo" followed by an optional param "int" can be compressed into one string:</p>
      * <ul>
-     *     <li>{@code foo [<int>]}</li>
+     * <li>{@code foo [<int>]}</li>
      * </ul>
      *
      * <p>The path to the specified node will <b>not</b> be prepended to the output, as there can theoretically be many
@@ -579,6 +580,22 @@ public class CommandDispatcher<S> {
         return getCompletionSuggestions(parse, parse.getReader().getTotalLength());
     }
 
+    /**
+     * Gets suggestions for a parsed input string on what comes next, checking from the given cursor position.
+     *
+     * <p>As it is ultimately up to custom argument types to provide suggestions, it may be an asynchronous operation,
+     * for example getting in-game data or player names etc. As such, this method returns a future and no guarantees
+     * are made to when or how the future completes.</p>
+     *
+     * <p>The suggestions provided will be in the context of the end of the parsed input string, but may suggest
+     * new or replacement strings for earlier in the input string. For example, if the end of the string was
+     * {@code foobar} but an argument preferred it to be {@code minecraft:foobar}, it will suggest a replacement for that
+     * whole segment of the input.</p>
+     *
+     * @param parse the result of a {@link #parse(StringReader, Object)}
+     * @param cursor the index to fetch suggestions for
+     * @return a future that will eventually resolve into a {@link Suggestions} object
+     */
     public CompletableFuture<Suggestions> getCompletionSuggestions(final ParseResults<S> parse, int cursor) {
         final CommandContextBuilder<S> context = parse.getContext();
 
--- a/src/main/java/com/mojang/brigadier/ImmutableStringReader.java
+++ b/src/main/java/com/mojang/brigadier/ImmutableStringReader.java
@@ -3,24 +3,82 @@
 
 package com.mojang.brigadier;
 
+/**
+ * Basically a string reader that can not be modified.
+ */
 public interface ImmutableStringReader {
+    /**
+     * Returns the full input string.
+     *
+     * @return the full input string
+     */
     String getString();
 
+    /**
+     * Returns the remaining length, so the distance from the cursor to the end of the input.
+     *
+     * @return the remaining length, so the distance from the cursor to the end of the input
+     */
     int getRemainingLength();
 
+    /**
+     * Returns the total length of the input.
+     *
+     * @return the total length of the input
+     */
     int getTotalLength();
 
+    /**
+     * Returns the current cursor position.
+     *
+     * @return the current cursor position.
+     */
     int getCursor();
 
+    /**
+     * Returns the part of the input that was already read.
+     *
+     * @return the part of the input that was already read
+     */
     String getRead();
 
+    /**
+     * Returns the part of the input that was not yet read.
+     *
+     * @return the part of the input that was not yet read
+     */
     String getRemaining();
 
+    /**
+     * Checks if the reader has enough input to read {@code length} more characters.
+     *
+     * @param length the amount of characters to read
+     * @return true of the reader has enough input to read {@code length} more characters.
+     */
     boolean canRead(int length);
 
+    /**
+     * Checks if the reader can read at least one more character.
+     *
+     * @return true if the reader can read at least one more character
+     * @see #canRead(int)
+     */
     boolean canRead();
 
+    /**
+     * Returns the next character but without consuming it (so the cursor stays at its current position).
+     *
+     * @return the next character
+     * @see #peek(int)
+     */
     char peek();
 
+    /**
+     * Returns the character {@code offset} places from now without consuming it (so the cursor stays at its current
+     * position).
+     *
+     * @param offset the offset of the character to peek at
+     * @return the character {@code offset} places from now
+     */
     char peek(int offset);
 }
--- a/src/main/java/com/mojang/brigadier/StringReader.java
+++ b/src/main/java/com/mojang/brigadier/StringReader.java
@@ -5,6 +5,9 @@ package com.mojang.brigadier;
 
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 
+/**
+ * A convenient wrapper around a string that allows you to read and iterate through it.
+ */
 public class StringReader implements ImmutableStringReader {
     private static final char SYNTAX_ESCAPE = '\\';
     private static final char SYNTAX_QUOTE = '"';
@@ -26,6 +29,11 @@ public class StringReader implements ImmutableStringReader {
         return string;
     }
 
+    /**
+     * Sets the cursor index position.
+     *
+     * @param cursor the new curso position
+     */
     public void setCursor(final int cursor) {
         this.cursor = cursor;
     }
@@ -75,24 +83,51 @@ public class StringReader implements ImmutableStringReader {
         return string.charAt(cursor + offset);
     }
 
+    /**
+     * Reads the next character.
+     * <p>
+     * Same as {@link #peek()}, but also consumes the character.
+     *
+     * @return the read character
+     */
     public char read() {
         return string.charAt(cursor++);
     }
 
+    /**
+     * Skips a single character.
+     */
     public void skip() {
         cursor++;
     }
 
+    /**
+     * Checks if the character is allowed in a number.
+     *
+     * @param c the character to check
+     * @return true if the character is allowed to appear in a number
+     */
     public static boolean isAllowedNumber(final char c) {
         return c >= '0' && c <= '9' || c == '.' || c == '-';
     }
 
+    /**
+     * Skips all following whitespace characters as determined by {@link Character#isWhitespace(char)}.
+     */
     public void skipWhitespace() {
         while (canRead() && Character.isWhitespace(peek())) {
             skip();
         }
     }
 
+    /**
+     * Reads an integer.
+     * <p>
+     * The integer may only contain characters that match {@link #isAllowedNumber(char)}.
+     *
+     * @return the read integer
+     * @throws CommandSyntaxException if the command is no proper int
+     */
     public int readInt() throws CommandSyntaxException {
         final int start = cursor;
         while (canRead() && isAllowedNumber(peek())) {
@@ -110,6 +145,14 @@ public class StringReader implements ImmutableStringReader {
         }
     }
 
+    /**
+     * Reads a long.
+     * <p>
+     * The long may only contain characters that match {@link #isAllowedNumber(char)}.
+     *
+     * @return the read long
+     * @throws CommandSyntaxException if the command is no proper long
+     */
     public long readLong() throws CommandSyntaxException {
         final int start = cursor;
         while (canRead() && isAllowedNumber(peek())) {
@@ -127,6 +170,14 @@ public class StringReader implements ImmutableStringReader {
         }
     }
 
+    /**
+     * Reads a double.
+     * <p>
+     * The double may only contain characters that match {@link #isAllowedNumber(char)}.
+     *
+     * @return the read double
+     * @throws CommandSyntaxException if the command is no proper int
+     */
     public double readDouble() throws CommandSyntaxException {
         final int start = cursor;
         while (canRead() && isAllowedNumber(peek())) {
@@ -144,6 +195,14 @@ public class StringReader implements ImmutableStringReader {
         }
     }
 
+    /**
+     * Reads a float.
+     * <p>
+     * The float may only contain characters that match {@link #isAllowedNumber(char)}.
+     *
+     * @return the read float
+     * @throws CommandSyntaxException if the command is no proper int
+     */
     public float readFloat() throws CommandSyntaxException {
         final int start = cursor;
         while (canRead() && isAllowedNumber(peek())) {
@@ -161,14 +220,25 @@ public class StringReader implements ImmutableStringReader {
         }
     }
 
+    /**
+     * Checks if a character is allowed in an unquoted string or needs to be escaped.
+     *
+     * @param c the character to check
+     * @return true if the character is allowed to appear in an unquoted string
+     */
     public static boolean isAllowedInUnquotedString(final char c) {
         return c >= '0' && c <= '9'
-            || c >= 'A' && c <= 'Z'
-            || c >= 'a' && c <= 'z'
-            || c == '_' || c == '-'
-            || c == '.' || c == '+';
+                || c >= 'A' && c <= 'Z'
+                || c >= 'a' && c <= 'z'
+                || c == '_' || c == '-'
+                || c == '.' || c == '+';
     }
 
+    /**
+     * Reads an unquoted string, {@literal i.e.} for as long as {@link #isAllowedInUnquotedString(char)} returns true.
+     *
+     * @return the read string
+     */
     public String readUnquotedString() {
         final int start = cursor;
         while (canRead() && isAllowedInUnquotedString(peek())) {
@@ -177,6 +247,13 @@ public class StringReader implements ImmutableStringReader {
         return string.substring(start, cursor);
     }
 
+    /**
+     * Returns a quoted string.
+     *
+     * @return a quoted string or an empty String, if {@link #canRead} is false
+     * @throws CommandSyntaxException if the next character is not a quote. an invalid escape character is encountered
+     * or the closing quote was not found
+     */
     public String readQuotedString() throws CommandSyntaxException {
         if (!canRead()) {
             return "";
@@ -208,6 +285,13 @@ public class StringReader implements ImmutableStringReader {
         throw CommandSyntaxException.BUILT_IN_EXCEPTIONS.readerExpectedEndOfQuote().createWithContext(this);
     }
 
+    /**
+     * Reads a string, deciding between {@link #readQuotedString}  and {@link #readUnquotedString} based on whether the
+     * next character is a quote.
+     *
+     * @return the read string
+     * @throws CommandSyntaxException if an error occurs parsing the string
+     */
     public String readString() throws CommandSyntaxException {
         if (canRead() && peek() == SYNTAX_QUOTE) {
             return readQuotedString();
@@ -216,6 +300,12 @@ public class StringReader implements ImmutableStringReader {
         }
     }
 
+    /**
+     * Reads a single boolean.
+     *
+     * @return the read boolean
+     * @throws CommandSyntaxException if the input was empty or the read word is not a valid boolean
+     */
     public boolean readBoolean() throws CommandSyntaxException {
         final int start = cursor;
         final String value = readString();
@@ -233,6 +323,13 @@ public class StringReader implements ImmutableStringReader {
         }
     }
 
+    /**
+     * Peel at the next char and consume it, if it is the expected character{@literal .} If not, throw an exception.
+     *
+     * @param c the character that is expected to occur next
+     * @throws CommandSyntaxException if the character was not the expected character or the end of the input was
+     * reached
+     */
     public void expect(final char c) throws CommandSyntaxException {
         if (!canRead() || peek() != c) {
             throw CommandSyntaxException.BUILT_IN_EXCEPTIONS.readerExpectedSymbol().createWithContext(this, String.valueOf(c));
--- a/src/main/java/com/mojang/brigadier/arguments/ArgumentType.java
+++ b/src/main/java/com/mojang/brigadier/arguments/ArgumentType.java
@@ -3,23 +3,65 @@
 
 package com.mojang.brigadier.arguments;
 
+import com.mojang.brigadier.CommandDispatcher;
 import com.mojang.brigadier.StringReader;
 import com.mojang.brigadier.context.CommandContext;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import com.mojang.brigadier.suggestion.Suggestions;
 import com.mojang.brigadier.suggestion.SuggestionsBuilder;
+import com.mojang.brigadier.tree.ArgumentCommandNode;
 
 import java.util.Collection;
 import java.util.Collections;
 import java.util.concurrent.CompletableFuture;
 
+/**
+ * An argument an {@link ArgumentCommandNode} can take as its argument.
+ * <p>
+ * As different commands need different arguments (string, integer, long, double,…), this interface provides a
+ * way to define parsers for each of them.
+ *
+ * @param <T> the type of the argument
+ */
 public interface ArgumentType<T> {
+
+    /**
+     * Parses the given {@link StringReader} to an instance of this argument's generic type.
+     *
+     * @param reader the read to read from
+     * @return the parsed argument. If this is an {@code ArgumentType<Integer>} this would return an Integer.
+     * @throws CommandSyntaxException if the argument is malformed
+     */
     T parse(StringReader reader) throws CommandSyntaxException;
 
+    /**
+     * Gets suggestions for a parsed input string on what comes next.
+     *
+     * <p>As it is ultimately up to custom argument types to provide suggestions, it may be an asynchronous operation,
+     * for example getting in-game data or player names etc. As such, this method returns a future and no guarantees
+     * are made to when or how the future completes.</p>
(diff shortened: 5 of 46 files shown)


NeunEinser:
I agree. A java doc is a much needed first step in the right direction. It's especially neccasary if more and more features get merged in, so that the javadoc can be kept up to date as a requirement for future pull requests.
Most of the stuff I looked at so far was looking good. Maybe some of the stuff could be more detailed. I already made a pull request to your branch with some minor stuff I've found.

I know, documentation is what no one likes doing, but I still hope that many more people in here give their feedback.

src/main/java/com/mojang/brigadier/context/CommandContextBuilder.java
```diff
@@ -67,6 +119,13 @@ public S getSource() {
         return this;
     }
```
NeunEinser: This method is in fact called in lib code. It is used in the parse -method of both `LiteralCommandNode` and `ArgumentCommandNode`
Also it does not "Overwrite" the parameters.

The `range` becomes the old range plus the new range. So if, in this particular CommandContext, the parsed range was 5, 10 and the given range is 10, 16, the new range will be 5, 16 as this is what was parsed in this context.

Both modifer and forks are always null/false (aka not set), unless it's the last node of a particular command. Attemting to add a node to a CommandNode with a redirect will throw an `IllegalStateException`, similar when trying to add a child node to a command node that has a redirect modifier.
Of course, since this is a public memeber, you could abuse this method in many ways, but it is really only intended to be used by `CommandNode#parse(StringReader, CommandContextBuilder)`. Maybe it should throw Exceptions if someone tries to do stupid stuff with it.

src/main/java/com/mojang/brigadier/CommandDispatcher.java
```diff
@@ -435,8 +436,8 @@ public int execute(final ParseResults<S> parse) throws CommandSyntaxException {
      * listed as multiple entries: the parent node, and the child nodes.
      * For example, a required literal "foo" followed by an optional param "int" will be two nodes:</p>
      * <ul>
-     *     <li>{@code foo}</li>
```
Marcono1234: Why did you unindent HTML elements?

I-Al-Istannen: Sigh. That was IntelliJs autoformatter. I surrounded the ones I wrote with formatter off tags to prevent that, but it seems I missed some. Thanks, I will restore that :)

Marcono1234:
1. I think when the `@return` text is the same as the method description, as it is the case for `ImmutableStringReader.java`, then only `@return` should be used. Otherwise it is more work to maintain and more to read despite adding no new information.

2. Why did you use `@literal` for some periods, e.g. https://github.com/Mojang/brigadier/pull/47/files#diff-c8e5c02eb71c3fa5ea2330d4f0c9267dR327 ?

3. Maybe it would be good to replace the Minecraft related documentation
    > getting in-game data or player names etc

    with something more generic or explain this a little bit more, since (if I understand it correctly) this project is intended as a general purpose library.

4. Not so important, but currently it is a little bit inconsistent whether a period is used at the end of `@param` and `@return`
5. `ArgumentBuilder.java` and similar classes often call nodes "commands". In my opinion this is misleading since nodes can have a command and I would not consider a node without command a "command". I did not add this as review comment everywhere.
6. There is no need for links to the enclosing class. For some of them I added review comments but not for all. These links are probably rather distracting. E.g.
```java
class A {
    /**
     * Creates new {@link A}.
     */
    A() {
        ...
    }
}
```


Sorry if the reviews are not helpful, annoying or unfriendly. I stopped before `SuggestionContext.java`.

src/main/java/com/mojang/brigadier/context/CommandContextBuilder.java
NeunEinser: Tbh, I don't know how useful the context builder in general is as a public member. Usually, when executing a command, you do not get access to the builder, only the context after building.
The only reason it's public is probably because it gets used by other packages internally. In fact, the only class that creates instances of this class is the CommandDispatcher.

I-Al-Istannen: Yea, it would have been best as an internal class, but that's not gonna happen in java.

I guess I will throw a disclaimer at the class doc, that this is an internal class and all bets are off.
But I think I will leave the javadoc on its members, to aid people mangling with the library.

src/main/java/com/mojang/brigadier/Command.java
```diff
@@ -6,9 +6,30 @@
 import com.mojang.brigadier.context.CommandContext;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 
+/**
+ * The functional interface actually representing a command to execute.
+ * <p>
```
kashike: See https://github.com/Mojang/brigadier/blob/master/src/main/java/com/mojang/brigadier/CommandDispatcher.java#L70 - wrap in `<p>` like that.

src/main/java/com/mojang/brigadier/ImmutableStringReader.java
```diff
@@ -3,24 +3,82 @@
 
 package com.mojang.brigadier;
 
+/**
+ * Basically a string reader that can not be modified.
```
Marcono1234: Should maybe explain the concept of the cursor here.

src/main/java/com/mojang/brigadier/Command.java
Marcono1234: Just as a note to anyone reading, this is not the only occurrence of `<p>` being used like this.

I-Al-Istannen: It is one out of 77 places where it is used like that, the only times where it is used differently is in CommandDispatcher. It isn't well formed HTML, but reads quite nice IMHO. I just forgot to adjust the existing ones.

I can adjust it to match the way it was done in CommandDispatcher instead though.

src/main/java/com/mojang/brigadier/StringReader.java
```diff
@@ -161,14 +220,25 @@ public float readFloat() throws CommandSyntaxException {
...
     public static boolean isAllowedInUnquotedString(final char c) {
         return c >= '0' && c <= '9'
-            || c >= 'A' && c <= 'Z'
-            || c >= 'a' && c <= 'z'
-            || c == '_' || c == '-'
-            || c == '.' || c == '+';
+                || c >= 'A' && c <= 'Z'
+                || c >= 'a' && c <= 'z'
+                || c == '_' || c == '-'
+                || c == '.' || c == '+';
```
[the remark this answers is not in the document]
I-Al-Istannen: No, the formatter profile was auto-inferred by IntelliJ and picked a continuation indent of 8 for some reason.

src/main/java/com/mojang/brigadier/StringReader.java
```diff
@@ -177,6 +247,13 @@ public String readUnquotedString() {
         return string.substring(start, cursor);
     }
 
+    /**
+     * Returns a quoted string.
```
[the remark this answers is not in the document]
I-Al-Istannen: Sounds good :)
Unfortunately the fields are private for some reason, so they would not appear in the generated javadoc (It would just place an empty String there), so I will literally duplicate them and hope I am not forcing the API to follow something it does not intend to keep. But quotes are hopefully unlikely to change ;)

src/main/java/com/mojang/brigadier/arguments/StringArgumentType.java
```diff
@@ -84,17 +142,39 @@ private static String escape(final String input) {
         return result.toString();
     }
 
+    /**
+     * The type of phrase that is matched by {@link StringArgumentType}.
+     */
     public enum StringType {
+        /**
+         * Matches a single word, optionally concatenated with underscores.
```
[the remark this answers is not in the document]
I-Al-Istannen: It is important though, as it means that an underscore does not constitute a word boundary. The word boundary is drawn by `StringReader#isAllowedInUnquotedString` currently. I constrained that further to just underscores, as I am not sure whether that part should really be part of the contract.

When thinking about it again, it probably should be. What do you think?

src/main/java/com/mojang/brigadier/arguments/StringArgumentType.java
```diff
@@ -84,17 +142,39 @@ private static String escape(final String input) {
...
+         * Can contain the following characters: <br>
+         * {@code [a-zA-Z0-9_\-.+]}
```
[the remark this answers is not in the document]
I-Al-Istannen: That kinda resolves the conversation above, I will link that method instead.

src/main/java/com/mojang/brigadier/builder/ArgumentBuilder.java
```diff
@@ -31,6 +56,12 @@ public T then(final ArgumentBuilder<S, ?> argument) {
         return getThis();
     }
 
+    /**
+     * Adds the given {@link CommandNode} as a child.
+     *
+     * @param argument the command node to add
+     * @return this object
+     */
```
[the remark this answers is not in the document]
I-Al-Istannen: Yea, that should be in the doc, I will add it.
I would however only document checked exceptions in the throws clause, not unchecked ones.

src/main/java/com/mojang/brigadier/builder/ArgumentBuilder.java
```diff
@@ -39,40 +70,116 @@ public T then(final CommandNode<S> argument) {
...
+     * @param command the {@link Command} to execute
+     * @return this object
+     */
     public T executes(final Command<S> command) {
         this.command = command;
         return getThis();
     }
 
+    /**
+     * Returns the {@link Command} the built {@link CommandNode} will execute.
```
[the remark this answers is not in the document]
I-Al-Istannen: I added it to the doc of the builder and the command node.

There does not seem to be any standard pulled in however ( like findbugs, intelliJ, google, JSR, ...) or my IDE doesn't suggest it.
I don't really want to make that decision, but it would be nice to have, yea.

src/main/java/com/mojang/brigadier/context/CommandContextBuilder.java
```diff
@@ -67,6 +119,13 @@ public S getSource() {
...
 
+    /**
+     * Creates a copy of this {@link CommandContextBuilder}.
+     * <p>
+     * Not a deep copy as Command, child, range and the arguments themselves are shared.
```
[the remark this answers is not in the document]
I-Al-Istannen: Ouch... Way to be stupid.

src/main/java/com/mojang/brigadier/context/CommandContextBuilder.java
I-Al-Istannen: Hm. It comes in handy if you add a new command node type. Not sure if doing that is expected though, or if it should only be done by the library itself.

I-Al-Istannen:
@Marcono1234 That was quite thorough and not annoying or unfriendly. Thank you very much!

I have applied most of it, except for a few where I wasn't sure.
You caught *a lot*, so again, thank you very much! :)

src/main/java/com/mojang/brigadier/arguments/StringArgumentType.java
```diff
@@ -147,9 +147,9 @@ private static String escape(final String input) {
      */
     public enum StringType {
         /**
-         * Matches a single word, optionally concatenated with underscores.
-         * Can contain the following characters: <br>
-         * {@code [a-zA-Z0-9_\-.+]}
+         * Matches a single word.
+         * Can contain the characters outline in {@link StringReader#isAllowedInUnquotedString(char)}.
```
Marcono1234: "outline" -> "outlined"?

src/main/java/com/mojang/brigadier/arguments/ArgumentType.java
```diff
@@ -49,18 +51,20 @@
      * @param context the context to get them for
      * @param builder the suggestions builder to add them to
      * @return a future that will eventually resolve into a {@link Suggestions} object
-     * @implSpec The default implementation simple returns an empty list.
+     * @implSpec The default implementation simple returns an empty suggestions object
```
Marcono1234: "simple returns" -> "simply returns"

src/main/java/com/mojang/brigadier/context/ParsedArgument.java
```diff
@@ -16,7 +16,7 @@
     private final T result;
 
     /**
-     * Creates a new ParsedArgument for a given string range with a  given value.
+     * Creates a new {@code ParsedArgument} for a given string range with a  given value.
```
Marcono1234: No `{@code ...}`. Otherwise it is inconsistent with other methods and constructors.

src/main/java/com/mojang/brigadier/Command.java
kashike: While `</p>` *can* be omitted, I personally find it nicer to format as done in `CommandDispatcher` - it is nice to see "it starts here and ends here" instead of just "it starts here and automatically ends... here".

src/main/java/com/mojang/brigadier/arguments/StringArgumentType.java
I-Al-Istannen: I just forgot to delete that one, sorry :/

And yes.

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -42,30 +49,73 @@ protected CommandNode(final Command<S> command, final Predicate<S> requirement,
         this.forks = forks;
     }
 
+    /**
+     * Returns the command to execute when executing this command.
```
Marcono1234:
```suggestion
     * Returns the command to execute when executing this command node.
```

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -42,30 +49,73 @@ protected CommandNode(final Command<S> command, final Predicate<S> requirement,
...
+     *
+     * @return the command to execute when executing this command or null if not set
```
Marcono1234:
```suggestion
     * @return the command to execute when executing this command node or null if not set
```

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -42,30 +49,73 @@ protected CommandNode(final Command<S> command, final Predicate<S> requirement,
...
+     */
     public Command<S> getCommand() {
         return command;
     }
 
+    /**
+     * Returns all child commands.
```
Marcono1234:
```suggestion
     * Returns all child command nodes.
```

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -42,30 +49,73 @@ protected CommandNode(final Command<S> command, final Predicate<S> requirement,
...
+     *
+     * @return all child commands
```
Marcono1234:
```suggestion
     * @return all child command nodes
```

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -42,30 +49,73 @@ protected CommandNode(final Command<S> command, final Predicate<S> requirement,
...
+     */
     public Collection<CommandNode<S>> getChildren() {
         return children.values();
     }
 
+    /**
+     * Returns a child with the given {@link #getName}.
+     *
+     * @param name the name of the child command
```
Marcono1234:
```suggestion
     * @param name the name of the child node
```

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -42,30 +49,73 @@ protected CommandNode(final Command<S> command, final Predicate<S> requirement,
...
+     * Returns the redirect modifier to apply when redirecting.
+     *
+     * @return the redirect modifier to apply when redirecting or null if none
+     */
     public RedirectModifier<S> getRedirectModifier() {
         return modifier;
     }
 
+    /**
+     * Checks whether the given command source can use this command.
```
Marcono1234:
```suggestion
     * Checks whether the given command source can use this command node.
```

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -42,30 +49,73 @@ protected CommandNode(final Command<S> command, final Predicate<S> requirement,
...
+     * <p>
+     * This just checks the {@link #getRequirement()} predicate, which could e.g. check for permissions.
+     *
+     * @param source the command source to check for
+     * @return true if the given command source can use this command
```
Marcono1234:
```suggestion
     * @return true if the given command source can use this command node
```

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -42,30 +49,73 @@ protected CommandNode(final Command<S> command, final Predicate<S> requirement,
...
     public boolean canUse(final S source) {
         return requirement.test(source);
     }
 
+    /**
+     * Adds a new child node to this command node.
+     * <p>
+     * This will replace commands with the same name.
+     *
+     * <strong>You are not allowed to add children to commands with a {@link #getRedirect()} target!</strong>
```
Marcono1234: Ideally an exception should be thrown here, though not part of this pull request

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -42,30 +49,73 @@ protected CommandNode(final Command<S> command, final Predicate<S> requirement,
...
+     * @param node the child command node to add
+     * @throws UnsupportedOperationException if you try to add a {@link RootCommandNode} to any other command
```
Marcono1234:
```suggestion
     * @throws UnsupportedOperationException if you try to add a {@link RootCommandNode} to any other node
```

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -42,30 +49,73 @@ protected CommandNode(final Command<S> command, final Predicate<S> requirement,
...
+     * @return true if the given command source can use this command
+     */
     public boolean canUse(final S source) {
         return requirement.test(source);
     }
 
+    /**
+     * Adds a new child node to this command node.
+     * <p>
+     * This will replace commands with the same name.
```
Marcono1234:
```suggestion
     * This will replace an existing node with the same name.
```

Though this is not entirely true, see #32, maybe the doc should refer to that issue or state that it tries to merge the nodes, however doing this is currently discouraged.

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -92,6 +142,14 @@ public void addChild(final CommandNode<S> node) {
         children = children.entrySet().stream().sorted(Map.Entry.comparingByValue()).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, LinkedHashMap::new));
     }
 
+    /**
+     * Tries to find ambiguities in the children of this command and recurses down.
```
Marcono1234:
```suggestion
     * Tries to find ambiguities in the children of this command node and recurses down.
```

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -117,6 +175,15 @@ public void findAmbiguities(final AmbiguityConsumer<S> consumer) {
         }
     }
 
+    /**
+     * Checks if the given input is valid for this command, i.e. if it is what the command expects.
```
Marcono1234:
```suggestion
     * Checks if the given input is valid for this command node, i.e. if it is what the node expects.
```

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -117,6 +175,15 @@ public void findAmbiguities(final AmbiguityConsumer<S> consumer) {
...
+     * <p>
+     * This is used to find ambiguities.
+     *
+     * @param input the input to check
+     * @return true if the given input is valid
+     * @see #findAmbiguities}
```
Marcono1234:
```suggestion
     * @see #findAmbiguities
```

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -137,22 +204,70 @@ public int hashCode() {
...
+     * Returns the requirement for this command, that is used by {@link #canUse(Object)}.
+     *
+     * @return the requirement for this command, that is used by {@link #canUse(Object)}
+     */
     public Predicate<S> getRequirement() {
         return requirement;
     }
 
+    /**
+     * Returns the name of this command.
```
Marcono1234: "command" -> "command node" / "node"; same for following other methods of this class

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -137,22 +204,70 @@ public int hashCode() {
...
+     * Creates a builder for this command.
+     *
+     * @return a builder for this command
+     */
     public abstract ArgumentBuilder<S, ?> createBuilder();
 
+    /**
+     * Returns a key for this command, that is used for sorting the commands
+     *
+     * @return a key for this command, that is used for sorting the output
```
Marcono1234: What is "the output"?

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -181,9 +296,22 @@ public int compareTo(final CommandNode<S> o) {
         return (o instanceof LiteralCommandNode) ? 1 : -1;
     }
 
+    /**
+     * Checks whether this command forks.
+     * <p>
+     * See {@link CommandDispatcher#execute(ParseResults)} for an explanation of what it does
```
Marcono1234:
```suggestion
     * See {@link CommandDispatcher#execute(ParseResults)} for an explanation of what it does.
```

src/main/java/com/mojang/brigadier/tree/LiteralCommandNode.java
```diff
@@ -27,6 +34,11 @@ public LiteralCommandNode(final String literal, final Command<S> command, final
         this.literal = literal;
     }
 
+    /**
+     * Returns the literal this command is triggered by.
```
Marcono1234:
```suggestion
     * Returns the literal this node represents.
```

Or maybe something else instead of "represents"

src/main/java/com/mojang/brigadier/tree/LiteralCommandNode.java
```diff
@@ -27,6 +34,11 @@ public LiteralCommandNode(final String literal, final Command<S> command, final
...
+     *
+     * @return the literal this command is triggered by
```
Marcono1234:
```suggestion
     * @return the literal this node represents
```

src/main/java/com/mojang/brigadier/tree/RootCommandNode.java
```diff
@@ -51,16 +85,32 @@ public boolean equals(final Object o) {
...
+     *
+     * @return an empty string
+     */
     @Override
     protected String getSortedKey() {
         return "";
     }
 
+    /**
+     * Returns an empty list
```
Marcono1234:
```suggestion
     * Returns an empty immutable collection
```

src/main/java/com/mojang/brigadier/tree/RootCommandNode.java
```diff
@@ -51,16 +85,32 @@ public boolean equals(final Object o) {
...
+     *
+     * @return an empty list
```
Marcono1234:
```suggestion
     * @return an empty immutable collection
```

src/main/java/com/mojang/brigadier/builder/ArgumentBuilder.java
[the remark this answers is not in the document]
Marcono1234: Yes you are right, it is better to wait for the devs to change that then.

src/main/java/com/mojang/brigadier/suggestion/Suggestion.java
```diff
@@ -70,21 +132,58 @@ public int hashCode() {
     @Override
     public String toString() {
         return "Suggestion{" +
-            "range=" + range +
-            ", text='" + text + '\'' +
-            ", tooltip='" + tooltip + '\'' +
-            '}';
+                "range=" + range +
```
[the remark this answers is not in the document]
I-Al-Istannen: No, same error as before. I didn't trust the formatter to reformat everything, so that was leftover.

src/main/java/com/mojang/brigadier/suggestion/Suggestions.java
```diff
@@ -80,6 +112,38 @@ public static Suggestions merge(final String command, final Collection<Suggestio
...
+     * the computed general range.
+     * <p>
+     * After all of that it sorts the results and returns them.
+     * <p>
+     * <br>Some examples:
+     * <pre>
+     *     Suggestion foo = new Suggestion(StringRange.between(0, 2), "foo");
+     *     Suggestion bar = new Suggestion(StringRange.at(2), "bar");
+     *
+     *     Suggestions suggestions = Suggestions.create("1234567", List.of(foo, bar));
```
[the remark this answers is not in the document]
I-Al-Istannen: More meant as an example, yea. List.of was just nice and intuitive, but I can change it, if you wish.

src/main/java/com/mojang/brigadier/exceptions/CommandSyntaxException.java
```diff
@@ -5,6 +5,11 @@
...
 import com.mojang.brigadier.Message;
 
+/**
+ * An exception for a syntax error that occurred while parsing or executing a command.
+ * <p>
+ * TODO: Why is this named syntax exception, when it can also occur when executing a command?
+ */
 public class CommandSyntaxException extends Exception {
     public static final int CONTEXT_AMOUNT = 10;
     public static boolean ENABLE_COMMAND_STACK_TRACES = true;
```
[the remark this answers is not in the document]
I-Al-Istannen: Yea, that should really be configurable IMHO, at least optionally. I will document them though, as they are indeed part of the exported API. I missed them before, thank you :)

src/main/java/com/mojang/brigadier/suggestion/Suggestions.java
```diff
@@ -80,6 +112,38 @@ public static Suggestions merge(final String command, final Collection<Suggestio
...
+     * in the suggestions.
+     * <p>
+     * It will also automatically call {@link Suggestion#expand} for every possible suggestion, using the command and
+     * the computed general range.
+     * <p>
+     * After all of that it sorts the results and returns them.
+     * <p>
+     * <br>Some examples:
+     * <pre>
+     *     Suggestion foo = new Suggestion(StringRange.between(0, 2), "foo");
```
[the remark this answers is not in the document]
I-Al-Istannen: I indented them to make it a bit more readable in the source, but I removed the indent. The code is still indented a bit, but by four spaces less.

src/main/java/com/mojang/brigadier/tree/CommandNode.java
I-Al-Istannen: I have added a reference to the issue as well as stating that the current behaviour is not yet fixed, contains bugs, some design decisions are not yet decided and the result is therefore undefined.

src/main/java/com/mojang/brigadier/suggestion/Suggestion.java
```diff
@@ -8,33 +8,95 @@
 
 import java.util.Objects;
 
+/**
+ * A suggestion that the user could incorporate into his command, which can be used for displaying information about
+ * possible next values.
```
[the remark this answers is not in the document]
Marcono1234: Or use the existing sentence but restructure it to first mention what it does "information about next values" and then how it can be used.

src/main/java/com/mojang/brigadier/suggestion/Suggestions.java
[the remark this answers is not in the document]
Marcono1234: You can leave it like that if you want to, I don't care that much.

src/main/java/com/mojang/brigadier/exceptions/CommandSyntaxException.java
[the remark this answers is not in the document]
Marcono1234: `ENABLE_COMMAND_STACK_TRACES` also determines whether the stack trace should be writable, see [Throwable.fillInStackTrace()](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html#fillInStackTrace--)

src/main/java/com/mojang/brigadier/exceptions/CommandSyntaxException.java
```diff
@@ -5,8 +5,20 @@
...
+ * TODO: Why is this named syntax exception, when it can also occur when executing a command?
+ */
 public class CommandSyntaxException extends Exception {
+    /**
+     * The amount of characters from the input that should be displayed in the {@link #getContext()}.
+     */
     public static final int CONTEXT_AMOUNT = 10;
+    /**
+     * Enables suppression as defined in {@link Exception#Exception(String, Throwable, boolean, boolean)}, i.e.
+     * whether to ignore calls to {@link Exception#addSuppressed(Throwable)}
```
Marcono1234:
```suggestion
     * whether to ignore calls to {@link Exception#addSuppressed(Throwable)}.
```

src/main/java/com/mojang/brigadier/suggestion/Suggestion.java
```diff
@@ -8,33 +8,97 @@
 
 import java.util.Objects;
 
+/**
+ * Represents a suggestion which for a possible next value as well as an optional tooltip.
```
Marcono1234:
```suggestion
 * Represents a suggestion for a possible next value as well as an optional tooltip.
```

src/main/java/com/mojang/brigadier/suggestion/Suggestion.java
```diff
@@ -76,15 +140,66 @@ public String toString() {
...
+     *     </li>
+     *     <li>
+     *         {@code expand("123", StringRange.at(1)}:
+     *         <br>{@code a----1bcdefghi}, as the passed range (1) was taken and one character from the command was
+     *         used to fill up the index 1, as the range of the original suggestion was 0
+     *     </li>
+     *     <li>
+     *         {@code expand("123", StringRange.between(1, 3)}:
+     *         <br>{@code a----123defghi}, as the passed range (1,3) was taken and 3 characters from the command were
+     *         used to fill up the indices 1, 2 and 3, as the range of the original suggestion was 0. it is then inserted
```
Marcono1234:
```suggestion
     *         used to fill up the indices 1, 2 and 3, as the range of the original suggestion was 0. It is then inserted
```

I-Al-Istannen:
Thank you for the previous ones! I will go through all of it again when I have more time (quite busy with uni during the week atm) and see what I find. I apologize that the docs had a lot of typos and some inaccuracies, so I greatly appreciate the time you poured into ironing them out!

NJDaeger:
Is there any progress on this?? I'd love to see more docs.

PR #50, opened 2018-12-14
Add eclipse files to .gitignore.

JimmyHoke:
This pull request adds eclipse files to .gitignore so that contributers using eclipse can eality use the repo without leaving eclipse specific files.  This pull request has also removed a few useless declarations to make the the code easier to read.  It is a very small change but should start a trend toward better perfomrance.


kashike:
They aren't "useless declarations", they're named so that you can easily tell what they are.

Pokechu22:
:+1: to including eclipse in .gitignore, though I'm not sure src/.gitignore is the right location.  `gradlew eclipse` generates the eclipse configuration in the root of the project and not in src... though it looks like that actually requires enabling the eclipse plugin in build.gradle, which currently isn't enabled (nor is the intellij one).

I don't think the inlining is necessary from a performance perspective; per [JLS 13.1](https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html#jls-13.1) paragraph 3 the compiler automatically inlines them.  From a style perspective I can see both sides (in this case, the characters are _somewhat_ self-documenting) so I don't really have an opinion on that.

JimmyHoke:
Yes but “\\” makes more sense than SYNTAX_ESCAPE.  Otherwise newer programmers may be confused.

PolarizedIons:
imho, this should have been 2 PRs, as it's pretty separate issues.

JimmyHoke:
I have updated this pull request to only include the eclipse .gitignore.  Thanks for your feedback.

sjx233:
The original `.gitignore` was [created by https://www.gitignore.io/](https://github.com/Mojang/brigadier/blob/447845ba89c1285cff0178afef082339c93c0ba5/.gitignore#L1), so why not use 
https://www.gitignore.io/api/java,gradle,eclipse,intellij?

NeunEinser:
Why did you close it?
Also, while you're at it, you might want to include netbeans as well for people using that one. (https://www.gitignore.io/api/java,gradle,eclipse,intellij,netbeans)

PR #54, opened 2019-04-26
Add recursive command that does not cause stack overflow

liach:
## Reason
This is to better support vanilla *Minecraft* functions so that they can have a correct return result when they are nested.

This can also allow recursive commands in brigadier without concern for java heap space.

## Bugs
https://bugs.mojang.com/browse/MC-135636
https://bugs.mojang.com/browse/MC-148612

These bugs can have ugly workarounds by calling the result consumer twice (See https://github.com/liachmodded/MC-126946-128565/commit/b42d0455329408c2e2156d913be4ca2fd4255f8a#diff-e2109301c280d1578b4527aec48c2676R52), but this is the correct fix without erroneously calling result consumer multiple times.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,7 @@ dependencies {
     testCompile 'org.mockito:mockito-core:1.10.19'
     testCompile 'com.google.guava:guava-testlib:26.0-jre'
     testCompile 'org.openjdk.jmh:jmh-core:1.21'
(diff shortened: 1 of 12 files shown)


liach:
Shall close this one and create another one with a more comprehensive system such as max length limit. The tests for result consumer and in order, etc. will be moved to a new pull request.

PR #55, opened 2019-05-02
Use Locale.ENGLISH for case folding

Pr0methean:
`String.toLowerCase` gives different results depending on the server JVM's default locale (see http://lotusnotus.com/lotusnotus_en.nsf/dx/dotless-i-tolowercase-and-touppercase-functions-use-responsibly.htm), so for example "KICK" wouldn't become "kick" if the server locale was Turkish. Explicitly using English case folding makes the behavior consistent regardless of server locale.

### DIFF
--- a/src/main/java/com/mojang/brigadier/tree/LiteralCommandNode.java
+++ b/src/main/java/com/mojang/brigadier/tree/LiteralCommandNode.java
@@ -16,6 +16,7 @@ import com.mojang.brigadier.suggestion.SuggestionsBuilder;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Locale;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Predicate;
 
@@ -24,7 +25,7 @@ public class LiteralCommandNode<S> extends CommandNode<S> {
 
     public LiteralCommandNode(final String literal, final Command<S> command, final Predicate<S> requirement, final CommandNode<S> redirect, final RedirectModifier<S> modifier, final boolean forks) {
         super(command, requirement, redirect, modifier, forks);
-        this.literal = literal;
+        this.literal = literal.toLowerCase(Locale.ENGLISH);
     }
 
     public String getLiteral() {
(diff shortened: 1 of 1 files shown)


Pokechu22:
Good idea, though if I recall correctly the main Minecraft codebase uses `Locale.ROOT` instead of `Locale.ENGLISH` for this purpose.

src/main/java/com/mojang/brigadier/tree/LiteralCommandNode.java
```diff
@@ -24,7 +25,7 @@
 
     public LiteralCommandNode(final String literal, final Command<S> command, final Predicate<S> requirement, final CommandNode<S> redirect, final RedirectModifier<S> modifier, final boolean forks) {
         super(command, requirement, redirect, modifier, forks);
-        this.literal = literal;
+        this.literal = literal.toLowerCase(Locale.ENGLISH);
```
liach: This one still needs to be switched to
```java
this.literal = literal.toLowerCase(Locale.ROOT);
```

Pr0methean: Done.

kashike (dismissed):

liach (dismissed):
Should have covered all usages correctly.

Pr0methean:
Yeah, I ran grep over the code to check for other uses of toLowerCase, toUpperCase and equalsIgnoreCase.

ChloeDawn (dismissed):

SPGoding:
I don't think it's ok to convert literal to lowercase in the constructor, because not all literals in Minecraft are lowercased, e.g. literal `announceAdvancements` in command `/gamerule`.

Pr0methean:
@SPGoding Those literals aren't case-sensitive anyway, so the only place where it makes a difference is in the autocomplete.

liach:
It's 1.16 update time, @Dinnerbone please reconsider looking into this (given fry is looking into DFU fixes).

liach:
This is now superseded by #86

Marcono1234:
> This is now superseded by #86

The only difference seems to be that this pull request also changes it so literals are parsed case-sensitively [here](https://github.com/Mojang/brigadier/pull/55/files#diff-d0001400ba816f0e4c1fae3df98645de141c05e8c17430966e76fb570817a716R56), though that might be an undesired behavior change?

PR #56, opened 2019-06-26
Allow Russian letters in commands

xxDark:


Pokechu22:
I feel like it might be better to use something like [`Character.isAlphabetic`](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isAlphabetic-int-) (or `isLetter` or `isLetterOrDigit`, or one of the identifier-related methods).  Hardcoding individual languages seems like it might get messy quickly.

src/main/java/com/mojang/brigadier/builder/ArgumentBuilder.java
```diff
@@ -9,18 +9,26 @@
...
-    private Predicate<S> requirement = s -> true;
+    private final List<Predicate<S>> requirements = new ArrayList<>(1);
+    private final List<Predicate<S>> requirementsView;
     private CommandNode<S> target;
     private RedirectModifier<S> modifier = null;
     private boolean forks;
 
+    protected ArgumentBuilder() {
+        requirementsView = Collections.unmodifiableList(requirements);
+        requirements.add(__ -> true);
```
NeunEinser: This should not be needed anymore, in fact this is why `testCreateBuilder` of ArgumentCommandNode and LiteralCommand node fails (createBuilder will initialize a new builder with this default requirement and add the requirement from the node again, making the builder not equal to the node)

Since you have the same lambda as starting value for the reduce method over there, it should be fine to remove.
https://github.com/Mojang/brigadier/blob/66ab4d6ec9f3eb250ae53d0278ed65a267453aba/src/main/java/com/mojang/brigadier/tree/CommandNode.java#L46

NeunEinser: This should not be needed anymore, in fact this is why `testCreateBuilder` of `ArgumentCommandNode` and `LiteralCommandNode` fails (createBuilder will initialize a new builder with this default requirement and add the requirement from the node again, making the builder not equal to the node)

Since you have the same lambda as starting value for the reduce method over there, it should be fine to remove.
https://github.com/Mojang/brigadier/blob/66ab4d6ec9f3eb250ae53d0278ed65a267453aba/src/main/java/com/mojang/brigadier/tree/CommandNode.java#L46

src/main/java/com/mojang/brigadier/builder/ArgumentBuilder.java
```diff
@@ -53,12 +61,22 @@ public T executes(final Command<S> command) {
...
 
+    public T requiresAll(final Collection<Predicate<S>> requirements) {
+        this.requirements.addAll(requirements);
+        return getThis();
+    }
+
+    @Deprecated
     public Predicate<S> getRequirement() {
-        return requirement;
+        return requirements.get(0);
```
liach: Should rather return a combined view of requirements here.

src/main/java/com/mojang/brigadier/builder/ArgumentBuilder.java
```diff
@@ -9,18 +9,26 @@
...
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
 import java.util.function.Predicate;
 
 public abstract class ArgumentBuilder<S, T extends ArgumentBuilder<S, T>> {
     private final RootCommandNode<S> arguments = new RootCommandNode<>();
     private Command<S> command;
-    private Predicate<S> requirement = s -> true;
+    private final List<Predicate<S>> requirements = new ArrayList<>(1);
```
liach: Why do you need a list? This does not need an order.

build.gradle
```diff
@@ -38,6 +37,9 @@ dependencies {
     annotationProcessor 'org.openjdk.jmh:jmh-generator-annprocess:1.21'
 }
 
+compileJava.options.encoding = "UTF-8"
+compileTestJava.options.encoding = "UTF-8"
```
liach:
```gradle
tasks.withType(JavaCompile) {
    options.encoding = "UTF-8"
}
```

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -127,18 +131,21 @@ public boolean equals(final Object o) {
...
 
     @Override
     public int hashCode() {
         return 31 * children.hashCode() + (command != null ? command.hashCode() : 0);
     }
 
+    @Deprecated
     public Predicate<S> getRequirement() {
-        return requirement;
+        return requirements.get(0);
```
liach: Return the combined one here

src/test/java/com/mojang/brigadier/tree/ArgumentCommandNodeTest.java
```diff
@@ -92,7 +92,7 @@ public void testCreateBuilder() throws Exception {
         final RequiredArgumentBuilder<Object, Integer> builder = node.createBuilder();
         assertThat(builder.getName(), is(node.getName()));
         assertThat(builder.getType(), is(node.getType()));
-        assertThat(builder.getRequirement(), is(node.getRequirement()));
+        assertThat(builder.getRequirements(), is(node.getRequirements()));
```
liach: For comparing collections, I think you are supposed to use `containsInAnyOrder` instead of `is`.

src/test/java/com/mojang/brigadier/tree/LiteralCommandNodeTest.java
```diff
@@ -130,7 +130,7 @@ public void testEquals() throws Exception {
     public void testCreateBuilder() throws Exception {
         final LiteralArgumentBuilder<Object> builder = node.createBuilder();
         assertThat(builder.getLiteral(), is(node.getLiteral()));
-        assertThat(builder.getRequirement(), is(node.getRequirement()));
+        assertThat(builder.getRequirements(), is(node.getRequirements()));
```
liach: Same as above.

src/main/java/com/mojang/brigadier/builder/ArgumentBuilder.java
```diff
@@ -53,12 +61,22 @@ public T executes(final Command<S> command) {
     }
 
     public T requires(final Predicate<S> requirement) {
-        this.requirement = requirement;
+        requirements.add(requirement);
         return getThis();
     }
 
+    public T requiresAll(final Collection<Predicate<S>> requirements) {
```
liach: Accept `Iterable` here for convenience.

src/main/java/com/mojang/brigadier/arguments/ShortArgumentType.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public static ShortArgumentType shortArg(final short min) {
+        return shortArg(min, Short.MAX_VALUE);
+    }
+
+    public static ShortArgumentType shortArg(final short min, final short max) {
+        return new ShortArgumentType(min, max);
+    }
+
+    public static int getShort(final CommandContext<?> context, final String name) {
```
liach: Why doesn't this return `short`?

src/main/java/com/mojang/brigadier/arguments/UuidArgumentType.java
```diff
@@ -0,0 +1,19 @@
...
+import com.mojang.brigadier.exceptions.CommandSyntaxException;
+
+import java.util.UUID;
+
+public class UuidArgumentType implements ArgumentType<UUID> {
+
+    public static UuidArgumentType uuid() {
+        return new UuidArgumentType();
+    }
+
```
liach: `getUuid` static method, private constructor suggested (as boolean argument type has them).

Also can consider suggesting `-` when necessary.

src/main/java/com/mojang/brigadier/builder/ArgumentBuilder.java
```diff
@@ -11,16 +11,25 @@
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.function.Predicate;
 
 public abstract class ArgumentBuilder<S, T extends ArgumentBuilder<S, T>> {
+    private static final Predicate<?> TRUE = __ -> true;
```
liach: Can move this in a static method like 
```java
private static <T> Predicate<T> alwaysTrue() {
    return t -> true;
}
```
This prevents unchecked warnings.
Due to how java handles lambda like this, in fact, only one instance of the predicate will be ever created (instead of creating a new object every call)

src/main/java/com/mojang/brigadier/builder/ArgumentBuilder.java
```diff
@@ -11,16 +11,25 @@
...
     private Command<S> command;
-    private Predicate<S> requirement = s -> true;
+    private final Set<Predicate<S>> requirements = new HashSet<>(1);
+    private final Set<Predicate<S>> requirementsView;
     private CommandNode<S> target;
     private RedirectModifier<S> modifier = null;
     private boolean forks;
 
+    protected ArgumentBuilder() {
+        requirements.add((Predicate<S>) TRUE);
```
liach: No need for this here.

src/main/java/com/mojang/brigadier/builder/ArgumentBuilder.java
```diff
@@ -53,12 +62,22 @@ public T executes(final Command<S> command) {
...
 
+    public T requiresAll(final Iterable<Predicate<S>> requirements) {
+        requirements.forEach(this.requirements::add);
+        return getThis();
+    }
+
+    @Deprecated
     public Predicate<S> getRequirement() {
-        return requirement;
+        return requirements.stream().reduce(__ -> true, Predicate::and);
```
liach: If you have added the static method `alwaysTrue()`, you can consider calling it here.

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -28,18 +29,20 @@
...
 
-    protected CommandNode(final Command<S> command, final Predicate<S> requirement, final CommandNode<S> redirect, final RedirectModifier<S> modifier, final boolean forks) {
+    protected CommandNode(final Command<S> command, final Set<Predicate<S>> requirements, final CommandNode<S> redirect, final RedirectModifier<S> modifier, final boolean forks) {
         this.command = command;
-        this.requirement = requirement;
+        this.requirements = requirements;
         this.redirect = redirect;
         this.modifier = modifier;
         this.forks = forks;
+        this.combined = requirements.stream().reduce(__ -> true, Predicate::and);
```
liach: Given you use this, I guess that `alwaysTrue` utility method can be public

src/main/java/com/mojang/brigadier/tree/RootCommandNode.java
```diff
@@ -14,10 +14,12 @@
 import java.util.Collection;
 import java.util.Collections;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.Predicate;
 
 public class RootCommandNode<S> extends CommandNode<S> {
+    private static final Predicate<?> TRUE = __ -> true;
     public RootCommandNode() {
-        super(null, c -> true, null, s -> Collections.singleton(s.getSource()), false);
+        super(null, Collections.singleton((Predicate<S>) TRUE), null, s -> Collections.singleton(s.getSource()), false);
```
liach: `Collection.emptySet()` should suffice.

PR #57, opened 2019-07-24
Fix quoted strings being parsed as booleans

SPGoding:
`parseBoolean()` in `StringReader` parses specific quoted strings (e.g. `"true"`, `'false'`) as booleans, which is pretty weird.

This PR trys to fix this and adds a test for it.


liach:
It's 1.16 update time, @Dinnerbone please reconsider looking into this (given fry is looking into DFU fixes).

Issue #40, opened 2018-10-23
[Question] Attach a command description to each Node?


I-Al-Istannen:
## Problem
I want to provide a user interface with similiar information as this:
![grafik](https://user-images.githubusercontent.com/20284688/47382086-c67fbf00-d701-11e8-8ecb-9e6fd54b828b.png)

The leading name in dark aqua is the *name* of the command, the text in dark gray is the *description* of this command (i.e. what it does when invoked), the `<number>` is the amount of children and the red text is the usage.

Now on to the parts it displays and how I think you can find them in Brigardier:

|Feature | Implementation note|
|--------------------|--------------------------------------------------------------------------------------------------------------------------------------|
| Amount of children | A call to `getChildren` and its size|
| Usage| A call to `get(Smart)Usage`|
| Name | Not sure how to do this. Might have missed it or you can (ab)use some other feature for it |
| Description| I did not see any reference to anything like this in the source and Minecrafts's help command does not show anything like it either. |

Is anything like this planned or within the scope of the project? Are there any constraints it would have to fulfill, such as being localizable?

I apologize if I missed something that already allows this. I apologize if that is the case and would be happy, if the issue would be closed with a pointer to where I should look :)

ryantheleach:
In Sponge I don't believe we have a name, but we have an aliasing system, so the 'name' ends up being the primary alias of the command.

I've found it helpful to have both a short description, and a long description for commands when writing a help system.

The short description usually being limited to 1 line, for display in pagination / indexes.

The long description would be shown if the command was clicked on in the help system, and would be enough to fit on 1 chat display, at the largest size.

As long as commands had some concept of a short description, and a long description, then various mod frameworks and servers could list them however they need to.

I think it's a safe assumption that any new strings shown to players (that arn't directly commands themselves) would need to be localizable, and may be the reason why they havn't been added to commands already.

PR #61, opened 2019-11-30
Add description to command nodes

JesFot:
See #40 
For an eventual help command purpose, a description of a node may be useful.
In order to keep backward compatibility, I have duplicated some constructors.


liach:
Can't it become an info-displaying command that is bound to the node (like add this default command to all your tree nodes), so that for nodes that originally have no command like vanilla first literals (e.g. `advancement` `datapack` that reports unknown command on initial typing) will execute the dummy info command instead?

JesFot:
Command or node description should not be interpreted as a command.
To use your example: `/advancement` should only display that something is wrong with the command, not a description message about the command itself, `/help advancement` should do this.

AlekseyVideman:
Hey, guys. What's the news?

JesFot:
> Hey, guys. What's the news?

None, I honestly thought it would get a bit more traction on Mojang's side given that's a relatively little change...
I have no idea how they work on this and if they even saw my PR 🤷

I'll keep waiting, hoping for the best 😉

PR #62, opened 2020-03-31
Custom result type

ramidzkh:
Allows for a custom result type, other than just an integer

### DIFF
--- a/src/main/java/com/mojang/brigadier/AmbiguityConsumer.java
+++ b/src/main/java/com/mojang/brigadier/AmbiguityConsumer.java
@@ -8,6 +8,6 @@ import com.mojang.brigadier.tree.CommandNode;
 import java.util.Collection;
 
 @FunctionalInterface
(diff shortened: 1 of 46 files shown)


jearmstrong21:
I really like this - any word on whether this is a feature Mojang would want to use?

NeunEinser:
> I really like this - any word on whether this is a feature Mojang would want to use?

Mojang has not worked on PRs for like 1.5 years now, so nope

liach:
This is a great feature and it appears to work great. However this is not backward compatible and will need a major version bump most likely, but is worth it.

hYdos:
hmmm... wonder how we can get someone to review this

haykam821 (approved):

YTG1234 (approved):

Devan-Kerman (approved):

ramidzkh:
Looks like generic hell anyway

PR #66, opened 2020-04-12
Added unregister method

NgLoader:
Your now able to unregister a registed command

### DIFF
--- a/src/main/java/com/mojang/brigadier/CommandDispatcher.java
+++ b/src/main/java/com/mojang/brigadier/CommandDispatcher.java
@@ -98,6 +98,15 @@ public class CommandDispatcher<S> {
         return build;
     }
 
+	/**
+	* Utility method from unregister a command
+	*
+	* @param command node from register function
+	*/
+	public void unregister(final LiteralCommandNode<S> node) {
+		this.root.removeChild(node);
+	}
+
     /**
      * Sets a callback to be informed of the result of every command.
      *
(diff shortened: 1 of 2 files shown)


marcbal:
You should also overload your method with a String parameter (in case you lost the reference to your CommandNode)
```java
public void unregister(final String name)
```

```java
public void removeChild(final String name)
```

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -89,6 +89,14 @@ public void addChild(final CommandNode<S> node) {
         }
     }
 
+	public void removeChild(final CommandNode<S> node) {
```
MoritzR200, line 92: Nodes are uniquely identified by their name.
Therefore, and because it is the only information you actually use from the node, the parameter should probably be the 'nodeName' and not the node itself.

Especially since when using the intended Argument-builder syntax to declare a command you do not have a reference to the command node, but you do know the name.

MoritzR200 (changes requested):
I think adding the ability to remove command nodes is very important, since it allows more flexibility. For example this enables you to add and remove a command based on some condition, without having to reregister all the other commands (or use reflections).

I do however think, that the nodes should be removed based on their name and not on the *CommandNode* instance itself.

src/main/java/com/mojang/brigadier/tree/CommandNode.java
NgLoader: Okay, thats a good point

src/main/java/com/mojang/brigadier/CommandDispatcher.java
```diff
@@ -99,6 +99,15 @@ public LiteralCommandNode<S> register(final LiteralArgumentBuilder<S> command) {
         return build;
     }
 
+	/**
+	* Utility method from unregister a command
```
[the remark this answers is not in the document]
NgLoader: thanks

MoritzR200 (approved):
This does look a lot better than before. This will certainly expand the usability of this Framework significantly

PR #67, opened 2020-04-18
Feature: List element argument type implementation.

MairwunNx:
### *What do you think? I think it might be useful.*


liach:
Don't see a point of including it here. Isn't really useful within Minecraft and can be easily added by another library/application dependent on brigadier.

PR #68, opened 2020-05-09
Don't sort children on node addition to save on performance

aikar:
Everytime a child is added to the CommandNode, the children was sorted.
This action is extremely heavy with large node trees such as Minecraft.

From what I can see, sort order was not even needed by type since
the dispatcher parse checks argument and literal nodes instead.

Testing on /bossbar command seemed to have no impact to behavior.

Credit to https://github.com/PaperMC/Paper/pull/3172

Co-authored-by: virustotalop <virustotalop@gmail.com>

### DIFF
--- a/src/main/java/com/mojang/brigadier/tree/CommandNode.java
+++ b/src/main/java/com/mojang/brigadier/tree/CommandNode.java
@@ -20,12 +20,11 @@ import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
-import java.util.TreeMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Predicate;
 
 public abstract class CommandNode<S> implements Comparable<CommandNode<S>> {
-    private final Map<String, CommandNode<S>> children = new TreeMap<>();
+    private final Map<String, CommandNode<S>> children = new LinkedHashMap<>();
     private final Map<String, LiteralCommandNode<S>> literals = new LinkedHashMap<>();
     private final Map<String, ArgumentCommandNode<S, ?>> arguments = new LinkedHashMap<>();
     private final Predicate<S> requirement;
--- a/src/main/java/com/mojang/brigadier/tree/CommandNode.java
+++ b/src/main/java/com/mojang/brigadier/tree/CommandNode.java
@@ -20,14 +20,14 @@ import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
 
 public abstract class CommandNode<S> implements Comparable<CommandNode<S>> {
-    private Map<String, CommandNode<S>> children = new LinkedHashMap<>();
-    private Map<String, LiteralCommandNode<S>> literals = new LinkedHashMap<>();
-    private Map<String, ArgumentCommandNode<S, ?>> arguments = new LinkedHashMap<>();
+    private final Map<String, CommandNode<S>> children = new TreeMap<>();
+    private final Map<String, LiteralCommandNode<S>> literals = new LinkedHashMap<>();
+    private final Map<String, ArgumentCommandNode<S, ?>> arguments = new LinkedHashMap<>();
     private final Predicate<S> requirement;
     private final CommandNode<S> redirect;
     private final RedirectModifier<S> modifier;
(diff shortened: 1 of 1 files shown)


aikar:
This issue causes significant CPU hits every time a player logs in or changes world, so it provided a substantial improvement.

See numbers here: https://github.com/PaperMC/Paper/issues/3154#issuecomment-614493898

liach:
Hmm, it is good to improve load-time performance, but by switching to a tree map, how is the lookup performance at runtime? Minecraft doesn't add to the map any more at runtime, so the actual impact isn't that big.

Theoretically a tree map lookup shouldn't be slower than that of a linked hash map. Just need some profiling to show the actual performance at runtime (after all the command children are added).

aikar:
I don't believe that map is even queried at runtime, the other maps are used instead.

But regardless, worrying about get performance is a premature optimization, as this is not hot code. Tree map will absolutely be fast enough for something that runs so infrequently even if it was queried.

src/main/java/com/mojang/brigadier/tree/CommandNode.java
```diff
@@ -88,8 +88,6 @@ public void addChild(final CommandNode<S> node) {
                 arguments.put(node.getName(), (ArgumentCommandNode<S, ?>) node);
             }
         }
-
-        children = children.entrySet().stream().sorted(Map.Entry.comparingByValue()).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, LinkedHashMap::new));
```
liach, line 92: Notice that the map is comparing by value rather than by key, as seen here:

https://github.com/Mojang/brigadier/blob/9542f80de2685ffad8af653be87e04d508d81c65/src/main/java/com/mojang/brigadier/tree/CommandNode.java#L173-L180

With your patch, the map will mix and match the literal and argument nodes by the node name, which breaks the original contract of the `getChildren` method of returning literal nodes first in the resulting collection.

Imo a better solution would be keep a tree set and a hash map together.

aikar: Yes it is a behavior change, but one that doesn't seem to matter.

The dispatcher does not iterate this list. arguments and literals are stored in secondary maps and those maps are what is used.

If Mojang wants to make that order matter elsewhere, I'm going to leave it on them to solve that.

liach: I strongly oppose your suggestion. Imo you should just keep these nodes in a tree set instead (why do you need the string keys when you can just look up the string keys from the other two maps?) and the iteration order of these nodes is actually important as literal branches are of higher priority in case of conflict with argument branches.

aikar: No because addChild still looks up by name to merge. That would break stuff.

This change does not break anything as the only place its iterated is findAmbiguities which order is not important there.

I just re-verified this change and can confirm again that switching to a TreeMap here does not break any behavior.

The change as supplied solves the problem and does not break any behavior of the system, so that is all that is needed.

liach: My suggestion is to change `children` to a Tree*Set*. Look up by names will simply be look up to two maps instead. Also this is just an implementation detail, it doesn't hurt others.

Since `getChildren` is public api, what works for paper spigot does not necessarily work out well for other communities, like mojang or other modding communities. Changing this breaks users.

liach: Also another step back, since you've said messing up iteration order "does not break any behavior", you should've just removed this sorting logic in its entirety for the best result.

aikar: I meant iteration order of params vs types. Iteration order was not a defined contract on getChildren. It's a Collection<CommandNode<S>>, clearly indicating no expectations of order.

Regardless, Mojang can take it or leave it. I'm not willing to spend any more time debating this.

liach:
Conclusion of the debate above: 
With aikar's current setup, he should just keep this patch in his pr
https://github.com/Mojang/brigadier/pull/68/files#diff-2eb7953fa0b13555eda5df1df0ac9e82L91-L92

The rest of replacing linked hash map with tree map is pointless.

The new tree map does not preserve the iteration order that the linked hash map could guarantee as the tree map is comparing the string key instead of the command node value as done in linked hash map previously.

liach:
Oh, that section would be prioritizing literal command nodes over other types of command nodes (presumably only argument command nodes), and then sort each type of node by their sorted key.
https://github.com/Mojang/brigadier/blob/9542f80de2685ffad8af653be87e04d508d81c65/src/main/java/com/mojang/brigadier/tree/CommandNode.java#L173-L180
For instance, if we have subchildren `apple` `<banana>` `cactus` (`<>` denotes argument nodes), the original sort order would be `apple` `cactus` `<banana>` (literals first), while yours would be `apple` `<banana>` `cactus` (literal mixed with argument, sorted only by name)

slicedlime (approved):

boq (approved):

liach:
Hmm, something actually happens here! Interesting :eyes:

PR #70, opened 2020-05-21
Add collection support to ArgumentBuilder.then

Emalios:


liach:
Not much point since we already can chain `.then(a).then(b)` calls.

Emalios:
Yes, but how would you do after an unspecified list of arguments to add a
literal argument or to execute the command?

liach:
A for loop is not that hard.
```java
for (ArgumentBuilder<S, ?> argument : /* your arguments array or list */) {
  builder.then(argument);
}
```
What you have right now is not good imo as this will generate an array every time a single parameter is passed, harming the performance. Also this is not backward compatible with the existing method.

Issue #76, opened 2020-10-14
Pass context to ArgumentType.parse


MaryWeeb:
For a few of my projects I'm considering using Brigadier, however one limitation has been stopping me from using this: ArgumentType.parse is not context-aware. 

In many cases, the object to be returned requires context of where or by who the command was invoked. The simple solution would be to pass the source object to the parse method as second parameter, however this would cause breaking changes in the library. 

To solve this, I propose to give the `parse` method  two signatures: `parse(StringReader reader)` and `parse(StringReader reader, Object context)`, both with a default implementation which returns null. 
Sadly due to how the current arguments are implemented they do not supply the type of the context, but on a per-application basis this could be done with a manual cast in the parse() method. To handle both functions, the following could be done:
```java
    // ArgumentCommandNode.java
    @Override
    public void parse(final StringReader reader, final CommandContextBuilder<S> contextBuilder) throws CommandSyntaxException {
        final int start = reader.getCursor();
        final T result = type.parse(reader);
        if (result == null) {
            // The context-unaware method was not implemented, call the context-aware method
            result = type.parse(reader, contextBuilder.getSource())
        }
        if (result == null) {
            // Throw an error to avoid magic NullPointerExceptions in the user's code
            throw CommandSyntaxException("No implementation for parse on argument type!")  // Simple error as example, should be improved
        }
        final ParsedArgument<S, T> parsed = new ParsedArgument<>(start, reader.getCursor(), result);

        contextBuilder.withArgument(name, parsed);
        contextBuilder.withNode(this, parsed.getRange());
    }```

This way users can implement context-aware parse methods without breaking any existing code.

Terrain2:
Have you considered `ArgumentCommandNode.isValidInput()`? It does not *have* a context or source object and needs to call the `ArgumentType`'s parse method, how is that meant to work with this change?

https://github.com/Mojang/brigadier/blob/ff4286760e7aa1986edc5cf7b69717661d2cbd48/src/main/java/com/mojang/brigadier/tree/ArgumentCommandNode.java#L97-L106

if an `ArgumentType` requires a source object and uses the default implementation for no source, then any input for it will be considered invalid when searching for ambiguities (since most arguments won't have leading whitespace), and as such finding ambiguities will silently fail with the wrong results

[...]

PR #77, opened 2020-10-15
Pass context to ArgumentType.parse

martmists-gh:
This Pull Request implements the suggestion outlined in #76 

It has been implemented to be non-breaking unless an implemented argument type can return a nullable value. Since the CommandContext is not yet available at this point in time, we can pass the command source, from which the required context can usually be extracted.

### DIFF
--- a/src/main/java/com/mojang/brigadier/tree/ArgumentCommandNode.java
+++ b/src/main/java/com/mojang/brigadier/tree/ArgumentCommandNode.java
@@ -58,14 +58,15 @@ public class ArgumentCommandNode<S, T> extends CommandNode<S> {
     public void parse(final StringReader reader, final CommandContextBuilder<S> contextBuilder) throws CommandSyntaxException {
         final int start = reader.getCursor();
         T result = type.parse(reader);
+
         if (result == null) {
-            // The context-unaware method was not implemented, call the context-aware method
             result = type.parse(reader, contextBuilder.getSource());
         }
+
         if (result == null) {
-            // Throw an error to avoid magic NullPointerExceptions in the user's code
             throw CommandSyntaxException.BUILT_IN_EXCEPTIONS.dispatcherParseException().createWithContext(reader, "Argument Type " + type.getClass().getName() + " does not implement a 'parse' method!");
         }
+
         final ParsedArgument<S, T> parsed = new ParsedArgument<>(start, reader.getCursor(), result);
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/mojang/brigadier/tree/ArgumentCommandNode.java
```diff
@@ -56,7 +57,16 @@ public String getUsageText() {
     @Override
     public void parse(final StringReader reader, final CommandContextBuilder<S> contextBuilder) throws CommandSyntaxException {
         final int start = reader.getCursor();
-        final T result = type.parse(reader);
+        T result = type.parse(reader);
+
+        if (result == null) {
+            result = type.parse(reader, contextBuilder.getSource());
```
i509VCB, line 63: I think the more specific context should be attempted first, using the non-context method as a fallback.

i509VCB (changes requested):

src/main/java/com/mojang/brigadier/tree/ArgumentCommandNode.java
MaryWeeb: I implemented it this way to have the original purpose (i.e. minecraft) take the least complex path.

lenrik1589 (approved):

DietrichPaul (approved):
this is incredibly important to merge!!! well implemented, would need that😂

src/main/java/com/mojang/brigadier/tree/ArgumentCommandNode.java
i509VCB: Comment should be marked as unresolved, no discussion has occurred and Minecraft would still operate correctly under this case.

lenrik1589:
> this is incredibly important to merge!!! well implemented, would need that😂

i don't quite get if this is sarcasm or actual comment of approval?

DietrichPaul:
> > this is incredibly important to merge!!! well implemented, would need that😂
> 
> i don't quite get if this is sarcasm or actual comment of approval?

I am not joking i REALLY NEED THIS NOW PLS

lenrik1589:
you can just clobe PR repo and build it from source, doesn't take all that
long

martmists-gh:
I won't be continuing work on this feature, as there's a lot of complications mentioned in the issue linked in the first post.

PR #81, opened 2020-12-28
Ensure that the CompletableFuture returned by getCompletionSuggestions is always completed

dualspiral:
Prior to this commit, if one CompletableFuture failed to complete, the returned future would never be completed because thenRun required all futures to complete sucessfully. Code that relies on such a completion would then be left waiting ad infinitum, with the potential for a deadlock occurring if join or get is used on the future. In Minecraft, this results in client completions failing to be displayed if more than one node asks the server for results, because all but one are cancelled - an exceptional condition!

The solution I've gone for here is to just return the results of futures that completed sucessfully.

* Test added to verify behaviour when a node fails to return a completion.

### DIFF
--- a/src/main/java/com/mojang/brigadier/CommandDispatcher.java
+++ b/src/main/java/com/mojang/brigadier/CommandDispatcher.java
@@ -599,16 +599,15 @@ public class CommandDispatcher<S> {
             futures[i++] = future;
         }
 
-        final CompletableFuture<Suggestions> result = new CompletableFuture<>();
-        CompletableFuture.allOf(futures).thenRun(() -> {
+        return CompletableFuture.allOf(futures).handle((voidResult, exception) -> {
             final List<Suggestions> suggestions = new ArrayList<>();
             for (final CompletableFuture<Suggestions> future : futures) {
-                suggestions.add(future.join());
+                if (!future.isCompletedExceptionally()) {
+                    suggestions.add(future.join());
+                }
             }
-            result.complete(Suggestions.merge(fullInput, suggestions));
+            return Suggestions.merge(fullInput, suggestions);
         });
-
-        return result;
     }
 
     /**
--- a/src/test/java/com/mojang/brigadier/CommandDispatcherTest.java
+++ b/src/test/java/com/mojang/brigadier/CommandDispatcherTest.java
@@ -7,6 +7,8 @@ import com.google.common.collect.Lists;
 import com.mojang.brigadier.context.CommandContext;
 import com.mojang.brigadier.context.CommandContextBuilder;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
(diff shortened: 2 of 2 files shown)


hugmanrique:
I believe the returned `CompletableFuture` can still complete exceptionally (don't know if this is intended, as you say "just return the results of futures that completed sucessfully") if two or more futures are completed exceptionally (when one is still not completed and is joined in `handle`).

dualspiral:
> I believe the returned CompletableFuture can still complete exceptionally (don't know if this is intended, as you say "just return the results of futures that completed sucessfully") if two or more futures are completed exceptionally (when one is still not completed and is joined in handle). 

How can it? [allOf](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#allOf-java.util.concurrent.CompletableFuture...-) will only call handle when all futures are completed, it doesn't short circuit if an exception occurs. [I wrote this code to quickly check this](https://gist.github.com/dualspiral/a81570d85fad339e1d09d347af5e6dca), the exceptions return first and the future doesn't return at that point, it returns after everything has completed. In that handle block, I then check if the future does not contain an exception before I join on it. The futures array that is checked is the same one as passed in to allOf.

I'm not sure what your point about `thenApply` is to be honest - `thenApply` returns a new completion stage so if you're using `thenApply` elsewhere on the future that was returned it isn't going to affect that particular future.

2. is what is going on here, in order to get completions out of the engine. Most exceptions in this scenario will mean that a branch is unavailable to you anyway.

hugmanrique:
My bad, I couldn't find the non-shortcircuit claim on the javadoc since I forgot to read the @returns: _a new CompletableFuture that is completed when all of the given CompletableFutures complete_. I'm sorry for wasting your time.

liach (approved):

hugmanrique:
While cherry-picking this change to [Velocity's fork](https://github.com/velocitypowered/brigadier/tree/velocity-changes), the test fails with
```
java.lang.NullPointerException
	at java.util.TreeMap.getEntry(TreeMap.java:347)
	at java.util.TreeMap.get(TreeMap.java:278)
	at com.mojang.brigadier.tree.CommandNode.addChild(CommandNode.java:93)
	at com.mojang.brigadier.builder.ArgumentBuilder.then(ArgumentBuilder.java:42)
	at com.mojang.brigadier.CommandDispatcherTest.testCompletionWithErroredFutureReturnsCompletedFuture(CommandDispatcherTest.java:511)
```
There's some non-mocked method that Brigadier expects to return a non-null value. I replaced the mocked literal with an argument node whose custom `SuggestionProvider` returns the same failed future. https://github.com/VelocityPowered/brigadier/pull/9/commits/a90cdb7f4482fb535e95eb90c0b3d96fa3653ba2#diff-526d0bc30d02d884de7e3f851c3ec4a2cce9491301428d36a43fc53579ba779aR504

PR #82, opened 2021-02-14
Correct README punctuation

BeetMacol:
Add two missing dots and remove a whitespace before an exclamation mark


slicedlime (approved):

portlek (approved):
dead repo but anyway :D

PR #86, opened 2021-03-18
Cache lowercase versions of inputs and literals

slicedlime:
Cache lowercase versions of inputs and literals to avoid repeated calls to toLowerCase().

Thank you to Spottedleaf.

### DIFF
--- a/src/main/java/com/mojang/brigadier/CommandDispatcher.java
+++ b/src/main/java/com/mojang/brigadier/CommandDispatcher.java
@@ -20,6 +20,7 @@ import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
@@ -588,7 +589,7 @@ public class CommandDispatcher<S> {
 
         final String fullInput = parse.getReader().getString();
         final String truncatedInput = fullInput.substring(0, cursor);
-        final String truncatedInputLowerCase = truncatedInput.toLowerCase();
+        final String truncatedInputLowerCase = truncatedInput.toLowerCase(Locale.ROOT);
         @SuppressWarnings("unchecked") final CompletableFuture<Suggestions>[] futures = new CompletableFuture[parent.getChildren().size()];
         int i = 0;
         for (final CommandNode<S> node : parent.getChildren()) {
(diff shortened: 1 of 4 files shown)


portlek (changes requested):
toLowerCase(Locale.ENGLISH) or toLowerCase(Locale.ROOT)

src/main/java/com/mojang/brigadier/CommandDispatcher.java
```diff
@@ -588,12 +588,13 @@ private String getSmartUsage(final CommandNode<S> node, final S source, final bo
 
         final String fullInput = parse.getReader().getString();
         final String truncatedInput = fullInput.substring(0, cursor);
+        final String truncatedInputLowerCase = truncatedInput.toLowerCase();
```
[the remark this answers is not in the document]
liach: The turkish i is real :+1:

boq (approved):

boq (approved):

Pokechu22 (approved):

liach (approved):

src/main/java/com/mojang/brigadier/suggestion/SuggestionsBuilder.java
```diff
@@ -8,18 +8,27 @@
...
+    private final String remainingLowerCase;
     private final List<Suggestion> result = new ArrayList<>();
 
-    public SuggestionsBuilder(final String input, final int start) {
+    public SuggestionsBuilder(final String input, final String inputLowerCase, final int start) {
         this.input = input;
+        this.inputLowerCase = inputLowerCase;
         this.start = start;
         this.remaining = input.substring(start);
+        this.remainingLowerCase = inputLowerCase.substring(start);
```
Spottedleaf, line 27: It might not be the case that for calling toLowerCase on a string preserves the number of chars in the string, so using the start index there could cause the cached value to be offset. Better to use `this.remaining.toLowerCase(Locale.ROOT)`

liach: This is true https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#toUpperCase-java.util.Locale- but I wonder if there are other places in brigadier that depends on the assumption that a lowercased string has the same length as the original string and maps 1 to 1. If there is, only fixing this location doesn't solve the problem, and I recommend doing such a fix in a separate pull request instead.

SeargeDP (approved):

src/main/java/com/mojang/brigadier/suggestion/SuggestionsBuilder.java
Marcono1234: For Brigadier it looks like only `SuggestionsBuilder` (and code using it) performs case conversion. I have created #92 for the issue you described.

Issue #87, opened 2021-04-03
Argument node is not considered when literal is impermissible with same name


hugmanrique:
Consider a literal that has two children: a literal called `bar`, and an argument node of type `StringArgumentType.word()`. The literal is impermissible, i.e. the requirement returns `false`, while the argument node has a non-null `Command`. That is,
```java
subject.register(literal("foo")
    .then(
        literal("bar")
            .requires(source -> false)
    )
    .then(
        argument("argument", StringArgumentType.word())
            .executes(command)
    )
);
```

Now, executing `foo bar` will fail with `Incorrect argument for command at position 4: foo <--[HERE]` since the dispatcher doesn't consider the arguments nodes as a fallback when the literals requirements are not met. Is this by design?
[Failing test](https://gist.github.com/hugmanrique/38a46adb8a812bb2e2d692a277e03c7b)

PR #88, opened 2021-04-03
Consider arguments when literal is impermissible

hugmanrique:
Fixes https://github.com/Mojang/brigadier/issues/87


MoritzR200 (approved):
This is an great change that adds further possibilities to this Framework in an expected manner.
Effectively the changes only resolve an issue, that is unintuitive to the end-user and instead returns the expected result.

The changes themself are implemented very reasonably and Tests are added which both show why this change is needed and ensure, that nothing breaks because of these changes.

PR #89, opened 2021-04-25
Add redirects to the equality checks for nodes.

dualspiral:
They are added as reference equality checks rather than standard object equality checks because of a stack overflow that can occur if a redirect is to a parent of the current node. If two nodes are redirected to two different nodes that otherwise look the same, then they will not be equal, which is fine because if one of the nodes gains a new child then they won't be equal any more - so I think this is a sensible check to have.

If there is an alternative idea I'd be open to that too - as long as redirects are included in the equality contract somehow - this is just what I think is the most sensible thing to do.

Test included to demonstrate the change.

(See https://github.com/SpongePowered/Sponge/issues/3356 for more info on how this is affecting us.)

### DIFF
--- a/src/main/java/com/mojang/brigadier/tree/CommandNode.java
+++ b/src/main/java/com/mojang/brigadier/tree/CommandNode.java
@@ -128,13 +128,18 @@ public abstract class CommandNode<S> implements Comparable<CommandNode<S>> {
 
         if (!children.equals(that.children)) return false;
         if (command != null ? !command.equals(that.command) : that.command != null) return false;
-
-        return true;
+        // Because the primary use of a redirect is to be able to create commands that effectively loop,
+        // we can't do a standard equality check on them as we could very well end up with a stack overflow
+        // if a branch loops back to this node. However, they are still important here as two nodes that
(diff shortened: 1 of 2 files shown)


liach:
As far as I can see, the argument command nodes can be distinguished by name.
https://github.com/Mojang/brigadier/blob/cf754c4ef654160dca946889c11941634c5db3d5/src/main/java/com/mojang/brigadier/tree/ArgumentCommandNode.java#L105
Afaik, the name is only used in rendering in command suggestions, like that on the client.
:thinking: What prevents you from choosing different names for the command nodes to send to the client to distinguish these nodes?

dualspiral:
[Not for two literals that have the same name/literal but different redirects (think two literals in different commands that redirect back, such as an `-f` flag)](https://github.com/Mojang/brigadier/blob/cf754c4ef654160dca946889c11941634c5db3d5/src/main/java/com/mojang/brigadier/tree/LiteralCommandNode.java#L91).

chixvv1:
This fixes [MC-137380](https://bugs.mojang.com/browse/MC-137380)

g59jpq2w5w7 (approved):

PR #94, opened 2021-06-13
Cache several common argument types

astei:
This PR caches common argument types. This has the potential to reduce memory usage simply by reusing instances in the tree. I haven't measured the true impact but it ought to be significant.


portlek (approved):

peterix:

PR #95, opened 2021-06-13
Remove unnecessary literals map

astei:
Literals have some special handling in Brigadier, but as it turns out, it's also possible to derive what is and isn't a literal from the existing children. This reduces the amount of memory used to store a full 1.16.x command tree by about 18%.


portlek (approved):

xHyroM:
any updates?

PR #98, opened 2021-07-25
Add isNext(char) and isNext(Predicate<Character>) methods to ImmutableStringReader and StringReader

ErrorCraft:
I see myself and other people use `reader.canRead() && reader.peek() == c` (or `!reader.canRead() || reader.peek() != c`) all the time. This is very tedious to do, so this adds the `isAt(char)` method to the `ImmutableStringReader` interface and `StringReader` class.

### DIFF
--- a/src/main/java/com/mojang/brigadier/ImmutableStringReader.java
+++ b/src/main/java/com/mojang/brigadier/ImmutableStringReader.java
@@ -24,5 +24,5 @@ public interface ImmutableStringReader {
 
     char peek(int offset);
 
-    boolean isAt(char c);
+    boolean isNext(char c);
 }
--- a/src/main/java/com/mojang/brigadier/ImmutableStringReader.java
+++ b/src/main/java/com/mojang/brigadier/ImmutableStringReader.java
@@ -23,4 +23,6 @@ public interface ImmutableStringReader {
     char peek();
 
     char peek(int offset);
+
+    boolean isAt(char c);
 }
--- a/src/main/java/com/mojang/brigadier/ImmutableStringReader.java
+++ b/src/main/java/com/mojang/brigadier/ImmutableStringReader.java
@@ -3,6 +3,8 @@
 
 package com.mojang.brigadier;
 
+import java.util.function.Predicate;
+
 public interface ImmutableStringReader {
     String getString();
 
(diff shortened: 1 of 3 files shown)


Pokechu22:
I feel like `isAt` isn't the best name for this (`at` makes me think of an index, not a character); maybe `nextIs` or something like that would be better?

ErrorCraft:
Ah yeah, maybe something like `isNext` to be more consistent with the other method names like `canRead`? (It also reads a bit better imo)

ErrorCraft:
Changed the method name to `isNext`. I was also thinking of adding an `isNext` method that accepts a `Predicate<Character>` as a parameter to be able to check for more than one character. For example `canRead() && Character.isWhitespace(peek())` in the `skipWhitespace` method would become `isNext(Character::isWhitespace)`.

src/main/java/com/mojang/brigadier/StringReader.java
```diff
@@ -76,6 +78,16 @@ public char peek(final int offset) {
...
 
+    @Override
+    public boolean isNext(char c) {
+        return canRead() && peek() == c;
+    }
+
+    @Override
+    public boolean isNext(Predicate<Character> predicate) {
+        return canRead() && predicate.test(peek());
+    }
```
Marcono1234, line 89: Would it make sense to implement these as `default` methods of the interface instead?
(Though assuming no or only a few users implement their own `ImmutableStringReader` it likely makes no difference)

src/main/java/com/mojang/brigadier/ImmutableStringReader.java
```diff
@@ -23,4 +25,8 @@
     char peek();
 
     char peek(int offset);
+
+    boolean isNext(char c);
+
+    boolean isNext(Predicate<Character> predicate);
```
Marcono1234, line 31: What about using [`IntPredicate`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/function/IntPredicate.html) here? This would avoid boxing and unboxing. But maybe it does not matter much. And it might cause confusion, making users believe this handles code points (including >= `Character.MAX_VALUE`), so this would require renaming the parameter to for example `charPredicate`.

PR #123, opened 2022-12-02
Support for return values other than ints with backwards-compatability

SploxFox:
This PR allows commands to return any object. Because of Java's autoboxing, it is mostly backwards compatible with older code which returns ints. The only breaking changes are that old user code which references `Command<S>` must be changed to `Command<S, Integer>`, and the return value from command execution is now Object, so the type needs to be checked/cast to int. Please let me know if there are any other breaking changes. I didn't change the tests aside from adding the Integer generic as mentioned above, so everything else should be the same.

I think that many users of this library will find the minimal cost of refactoring worth it given the immense benefit of returning more than just ints.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,6 @@ hs_err_pid*
 
 
 repo/
+
+# vscode workspace settings
+.vscode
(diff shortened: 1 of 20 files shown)


M4ximumPizza (approved):
Looks good. I would say it would be approved.

SploxFox:
I changed behavior a bit compared to how it was previously in this PR:
* Default behavior for all numeric boxed primitives (int, short, double, etc.) is now to add them, not just ints. This makes the behavior a bit more consistent across numeric types.
* I added overloads for the `execute` method on CommandDispatcher which allow users to optionally specify a "base" commandResult for the other results to be combined into, ex. to get results in a list: `subject.execute(parsed, new ListCommandResult())`

PR #124, opened 2022-12-24
Fix wrong redirect behavior

zly2006:
## Overview

Brigadier set the variable `foundCommand` to true [HERE](https://github.com/Mojang/brigadier/blob/master/src/main/java/com/mojang/brigadier/CommandDispatcher.java#L234) even if no `Command<S>` was executed, and this cause a serious problem.

An example to this is in minecraft: if you execute the command `/xp query Steve`, you will not receive an error message from the server. This is because that `xp` is redirected from `experience`, and when brigadier found this command is redirected, it set `foundCommand` to true--even if nothing is actually executed.

## Ways to fix

I think the best way is to remove this wrong assignment. I tried in fabric to modify this code like below, and it works.

```java
package com.github.zly2006.testmod.mixin;

import com.mojang.brigadier.CommandDispatcher;
import com.mojang.brigadier.ParseResults;
import com.mojang.brigadier.context.CommandContext;
import org.spongepowered.asm.mixin.Mixin;
import org.spongepowered.asm.mixin.injection.At;
import org.spongepowered.asm.mixin.injection.Inject;
import org.spongepowered.asm.mixin.injection.ModifyVariable;
import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
import org.spongepowered.asm.mixin.injection.callback.LocalCapture;

import java.util.ArrayList;
import java.util.List;

@Mixin(value = CommandDispatcher.class, remap = false)
public class MixinCommandDispatcher {
    boolean modify = false;
    boolean valueBefore = false;
    @ModifyVariable(method = "execute(Lcom/mojang/brigadier/ParseResults;)I", at = @At(value = "STORE"), ordinal = 1)
    private boolean modify(boolean v) {
        if (modify) {
            modify = false;
            return valueBefore;
        }
        return v;
    }

    @Inject(method = "execute(Lcom/mojang/brigadier/ParseResults;)I", at = @At(value = "INVOKE", target = "Lcom/mojang/brigadier/context/CommandContext;hasNodes()Z"), locals = LocalCapture.CAPTURE_FAILSOFT)
    private void checkFix(ParseResults<?> parse, CallbackInfoReturnable<Integer> cir, int result, int successfulForks, boolean forked, boolean foundCommand, String command, CommandContext<?> original, List<?> contexts, ArrayList<?> next, int size, int i, CommandContext<?> context, CommandContext<?> child) {
        if (child.hasNodes()) {
            modify = true;
            valueBefore = foundCommand;
        }
        else {
            modify = false;
        }
    }
}
```

The result after my modification was tested in Minecraft 1.19. With this modification, `/xp query Steve` will throw a `CommandSyntaxException` as expected and an error message from the server was received.

### DIFF
--- a/src/main/java/com/mojang/brigadier/CommandDispatcher.java
+++ b/src/main/java/com/mojang/brigadier/CommandDispatcher.java
@@ -231,7 +231,6 @@ public class CommandDispatcher<S> {
                 if (child != null) {
                     forked |= context.isForked();
                     if (child.hasNodes()) {
-                        foundCommand = true;
                         final RedirectModifier<S> modifier = context.getRedirectModifier();
                         if (modifier == null) {
                             if (next == null) {


zly2006:
This bug is also reported at [JIRA](https://bugs.mojang.com/browse/MC-256419?jql=project%20%3D%20MC%20AND%20resolution%20%3D%20Unresolved%20AND%20text%20~%20%22redirected%20command%22%20ORDER%20BY%20updated%20DESC).

src/main/java/com/mojang/brigadier/CommandDispatcher.java
```diff
@@ -231,7 +231,6 @@ public int execute(final ParseResults<S> parse) throws CommandSyntaxException {
                 if (child != null) {
                     forked |= context.isForked();
                     if (child.hasNodes()) {
-                        foundCommand = true;
```
Gegy, line 234: Looks like this was here to make sure cases where redirect modifiers returned zero results don't give any error (e.g. with `/execute as` where the selector returns no results). `foundCommand` should probably still be set to true if `modifier.apply()` returns 0 results 🙂

Gegy (changes requested):
Thank you so much for this PR! 😊 Noticed a tiny issue, but should be a small change 🙂

zly2006:
Seems that <https://bugs.mojang.com/browse/MC-256419> and <https://bugs.mojang.com/browse/MC-165773> can be closed as fixed. @slicedlime @Gegy

slicedlime:
Not quite - they will be closed when a Minecraft version is released using the updated library.

PR #125, opened 2023-02-03
Avoid eager stringification of exception messages

intsuc:
This PR avoids stringifying exception messages eagerly to reduce the overhead on command failure.

### DIFF
--- a/src/main/java/com/mojang/brigadier/exceptions/CommandSyntaxException.java
+++ b/src/main/java/com/mojang/brigadier/exceptions/CommandSyntaxException.java
@@ -16,7 +16,7 @@ public class CommandSyntaxException extends Exception {
     private final int cursor;
 
     public CommandSyntaxException(final CommandExceptionType type, final Message message) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/mojang/brigadier/exceptions/CommandSyntaxException.java
```diff
@@ -16,15 +16,15 @@ public class CommandSyntaxException extends Exception {
     private final int cursor;
 
     public CommandSyntaxException(final CommandExceptionType type, final Message message) {
-        super(message.getString(), null, ENABLE_COMMAND_STACK_TRACES, ENABLE_COMMAND_STACK_TRACES);
+        super(null, null, ENABLE_COMMAND_STACK_TRACES, ENABLE_COMMAND_STACK_TRACES);
```
M4ximumPizza, line 19: For the 2nd null, could you delete it because its a duplicate? Same for line 27?

intsuc: The second `null` should not be a duplicate. Since `Exception` does not have such a constructor, if you delete the second `null`, it will result in an error.

PR #126, opened 2023-02-21
Fix for issue 109

rdipippo:
Literals can now be case insensitive, which can be specified with a flag in LiteralArgumentBuilder.

if the flag is true, the literal will be put into the **literals** hash in **CommandNode** with its key set to all lower case, and it will be gotten from the hash with a lowercased key.

When **CommandNode.parse** is called, it will check the flag and do a case insensitive check if necessary.

### DIFF
--- a/src/main/java/com/mojang/brigadier/builder/LiteralArgumentBuilder.java
+++ b/src/main/java/com/mojang/brigadier/builder/LiteralArgumentBuilder.java
@@ -9,14 +9,23 @@ import com.mojang.brigadier.tree.LiteralCommandNode;
 public class LiteralArgumentBuilder<S> extends ArgumentBuilder<S, LiteralArgumentBuilder<S>> {
     private final String literal;
 
-    private boolean isCaseInsensitive = false;
+    private final boolean ignoreCase;
 
-    protected LiteralArgumentBuilder(final String literal) {
+    public LiteralArgumentBuilder(final String literal) {
+        this(literal, false);
+    }
+
+    public LiteralArgumentBuilder(final String literal, final boolean ignoreCase) {
         this.literal = literal;
+        this.ignoreCase = ignoreCase;
     }
 
     public static <S> LiteralArgumentBuilder<S> literal(final String name) {
-        return new LiteralArgumentBuilder<>(name);
+        return new LiteralArgumentBuilder<>(name, false);
+    }
+    
+    public static <S> LiteralArgumentBuilder<S> literal(final String name, final boolean ignoreCase) {
+        return new LiteralArgumentBuilder<>(name, ignoreCase);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


src/main/java/com/mojang/brigadier/builder/LiteralArgumentBuilder.java
```diff
@@ -9,6 +9,8 @@
 public class LiteralArgumentBuilder<S> extends ArgumentBuilder<S, LiteralArgumentBuilder<S>> {
     private final String literal;
 
+    private boolean isCaseInsensitive = false;
```
Gamebuster19901, line 12: This should be final. Also rename the variable to `ignoreCase` as that is the convention.

src/main/java/com/mojang/brigadier/builder/LiteralArgumentBuilder.java
```diff
@@ -9,6 +9,8 @@
...
+
     protected LiteralArgumentBuilder(final String literal) {
```
Gamebuster19901, line 14: Add the additional `ignoreCase` parameter to the constructor.

Then create the method `public static <S> LiteralArgumentBuilder<S> literal(final String name, final boolean ignoreCase) {`

Then modify `literal(String)` to call the created method with `false` so that old code is still compatible.

src/main/java/com/mojang/brigadier/builder/LiteralArgumentBuilder.java
```diff
@@ -26,9 +28,16 @@ public String getLiteral() {
         return literal;
     }
 
+    public boolean isCaseInsensitive() { return isCaseInsensitive; }
```
Gamebuster19901, line 31: Rename to `ignoresCase()`

src/main/java/com/mojang/brigadier/tree/LiteralCommandNode.java
```diff
@@ -23,11 +23,14 @@
 public class LiteralCommandNode<S> extends CommandNode<S> {
     private final String literal;
     private final String literalLowerCase;
+    private boolean isCaseInsensitive = false;
```
Gamebuster19901, line 26: This should be final, also rename to `ignoreCase` as that is convention

src/main/java/com/mojang/brigadier/tree/LiteralCommandNode.java
```diff
@@ -39,6 +42,12 @@ public String getName() {
         return literal;
     }
 
+    public boolean isCaseInsensitive() { return isCaseInsensitive; }
```
Gamebuster19901, line 45: rename to ignoresCase()

src/main/java/com/mojang/brigadier/tree/LiteralCommandNode.java
```diff
@@ -39,6 +42,12 @@ public String getName() {
...
+
+    public void caseInsensitive(boolean isCaseInsensitive) {
```
Gamebuster19901, line 47: Remove this method, nodes should never be able to be modified after they have been created, otherwise it is possible to violate the guarantees stipulated in `CommandDispatcher#getPath()` and `CommandDispatcher#findNode()`

src/test/java/com/mojang/brigadier/CommandDispatcherTest.java
```diff
@@ -70,6 +81,14 @@ public void testCreateAndExecuteOffsetCommand() throws Exception {
...
     }
 
+    @Test
+    public void testCreateAndExecuteCaseInsensitiveOffsetCommand() throws Exception {
+        subject.register(literal("Foo").executes(command).caseInsensitive(true));
+
+        assertThat(subject.execute(inputWithOffset("/foo", 1), source), is(42));
+        verify(command).run(any(CommandContext.class));
+    }
+
```
Gamebuster19901, line 91: Additional tests for using ignoreCase as a command argument should be added. Currently this only these tests only test the base node.

Gamebuster19901 (changes requested):

Issue #103, opened 2021-11-07
[BUG] `StringReader.readUnquotedString()` does not support non-ASCII characters


Pante:
I'm the maintainer of [Chimera](https://github.com/Pante/chimera), a library that allows the Brigadier command framework to be used in Spigot plugins. Recently, a developer opened an interesting [issue](https://github.com/Pante/chimera/issues/329) in which they reported that argument parsing failed when non-ASCII characters were specified.

![Sample error](https://user-images.githubusercontent.com/71902968/140231051-71f966fb-74cc-481f-8188-4d3d7331db0e.png)

Looking into the issue, the cause of the argument parsing failing for non-ASCII characters can be traced to:
```java
    // StringArgumentType's parse() method
    @Override
    public String parse(final StringReader reader) throws CommandSyntaxException {
        if (type == StringType.GREEDY_PHRASE) {
            final String text = reader.getRemaining();
            reader.setCursor(reader.getTotalLength());
            return text;
        } else if (type == StringType.SINGLE_WORD) {
            return reader.readUnquotedString(); // <-- calls this
        } else {
            return reader.readString();
        }
    }
    
    // StringReader's readUnquotedString() method
    public String readUnquotedString() {
        final int start = cursor;
        while (canRead() && isAllowedInUnquotedString(peek())) { // <-- calls this
            skip();
        }
        return string.substring(start, cursor);
    }
    
    // Cause of failure
    public static boolean isAllowedInUnquotedString(final char c) {
        return c >= '0' && c <= '9'
            || c >= 'A' && c <= 'Z'
            || c >= 'a' && c <= 'z'
            || c == '_' || c == '-'
            || c == '.' || c == '+';
    }
```
This affects `StringArgumentType.word()`, `StringArgumentType.string()`, `StringReader.readString()` and `StringReader.readUnquotedString()` and any other dependencies on these classes/methods.

Maybe I'm missing some context but it seems weird to only allow a small subset of ASCII characters. In my opinion, this implementation of `StringReader.readUnquotedString()` is extremely wonky and should be refactored to support non-ASCII characters. 

In the interim, I decided to create a simple utility method that reads strings until a whitespace is encountered. It behaves similarly to `StringReader.readUnquotedString()` .

[Source](https://github.com/Pante/chimera/blob/master/commons/src/main/java/com/karuslabs/commons/command/Readers.java#L58)
```java
    public static String unquoted(StringReader reader) {
        var start = reader.getCursor();
[...]

Issue #44, opened 2018-11-07
[Bug] argument parser cannot parse string with ":"


DeStilleGast:
So while I was playing with this lib, I got it slowly working with a few commands except one.

I have tried every `StringArgumentType` (word and string) but it cannot read arguments with :

errors trown with `toast123:toast456`
`word` and `string`: ` throws: com.mojang.brigadier.exceptions.CommandSyntaxException: Expected whitespace to end one argument, but found trailing data at position 16: ...t toast123<--[HERE] `

Note: I can't use greedyString since I use more arguments after this argument.
It's late in the evening, may have overlooked something or I missed a class that I could use.

ryantheleach:
@DeStilleGast I don't think there's enough information here to help you effectively.

Can you post a code example of the command registration that you were attempting to parse, and the full command that you were entering?

As well as what you expected to happen, (clearly in this case, not throw errors) but it looks like this is only one small piece of the puzzle.

DeStilleGast:
@ryantheleach yeah sure, makes sense

the code for the command is:
```java
commandDispatcher.register(
	literal("checkit")
	.then(
		argument("server", string())
			.executes(c -> {
				String server = c.getArgument("server", String.class);
				int port = 25565;

				if(server.contains(":")) {
					String[] split = server.split(":");
					server = split[0];
					port = Integer.parseInt(split[1]);
				}

				checkWhitelisted(c.getSource(), server, port, c.getSource().getUsername());
				return 1;
			})
			.then(argument("user", string())
				.executes(c -> {
					String server = c.getArgument("server", String.class);
					int port = 25565;

					if(server.contains(":")) {
						String[] split = server.split(":");
						server = split[0];
						port = Integer.parseInt(split[1]);
					}

					checkWhitelisted(c.getSource(), server, port, c.getArgument("user", String.class));
					return 1;
				})))
);
```

and the command that I try to execute (command prefix removed): `checkit 127.0.0.1:25566`
what this command basicly should do is that is checks the database if a user is whitelisted or not (reply is done in the function itself)

Dinnerbone:
The best solution here would be a custom argument so you can validate input quicker and do the splitting of port at parse-time instead of execute-time.

If you need to use string, you'll have to quote it. `checkit "127.0.0.1:25566"` - unquoted strings can only contain a limited set of characters.

DeStilleGast:
but my issue is fixed, keep this bug open or can it be closed ?

Pharap:
[...]

PR #131, opened 2023-05-06
Add greedy charset option to string argument

Owen1212055:
This implements the concept of "greedy charsets" for brigadier string argument types.

Currently, word / quotable phrase arguments have a strict charset. For quotable phrase arguments, you can simply wrap the argument in quotes to allow these characters. However, for word arguments, you cannot use other characters at all. This is quite limiting, as even characters as simple as = are not allowed. This also blocks pretty much all international characters, making these argument types inconvenient to use.

In order to keep backward compatibility, a new "greedy charset" field has been added to the string argument types. This allows words/quoted strings to allow a much wider range of characters. StringReader also now has "greedy" methods, which will act as a way to have this functionality in other custom argument types if wanted. This means that any previous use of the argument/StringReader methods will use the legacy behavior.

### Word argument
Before:
image

Note: Word arguments can't even be quoted, so it's impossible to use these characters.
image
After:
(with greedy charset)
image

### String argument
Before:
image
image

After:
(with greedy charset)
image
image
Note, that we no longer need to wrap the argument in quotes.

### Fixes
Fixes https://github.com/Mojang/brigadier/issues/103
Fixes https://github.com/Mojang/brigadier/issues/44

### DIFF
--- a/src/main/java/com/mojang/brigadier/StringReader.java
+++ b/src/main/java/com/mojang/brigadier/StringReader.java
@@ -242,7 +242,7 @@ public class StringReader implements ImmutableStringReader {
         return readString(false);
     }
 
-    String readString(boolean asciiOnly) throws CommandSyntaxException {
+    private String readString(boolean asciiOnly) throws CommandSyntaxException {
         if (!canRead()) {
             return "";
         }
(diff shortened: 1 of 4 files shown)


StillLutto (approved):
Looks good, please accept this PR. This would be one of the most, if not the most, useful things for plugin developers using commands.

PR #135, opened 2023-08-11
Optimize parsing performance

intsuc:
### Overview

This optimization avoids unnecessary object allocations by not taking snapshots of the parsing state when the size of the relevant nodes is `1`. Parsing time becomes about 48.6 to 57.1% of the original.

#### Before Optimization f20bede62a516a11a468d27d3f1adde2085762bc

| Benchmark                            | Mode | Cnt |    Score |   Error | Units |
| :----------------------------------- | :--- | --: | -------: | ------: | :---- |
| `ParsingBenchmarks.parse_`           | avgt |  25 |  102.391 | ± 1.121 | ns/op |
| `ParsingBenchmarks.parse_a1i`        | avgt |  25 |  322.832 | ± 3.665 | ns/op |
| `ParsingBenchmarks.parse_c`          | avgt |  25 |  102.086 | ± 1.063 | ns/op |
| `ParsingBenchmarks.parse_k1i`        | avgt |  25 |  338.494 | ± 2.231 | ns/op |
| `ParsingBenchmarks.parse_l123456789` | avgt |  25 | 1014.364 | ± 9.041 | ns/op |

#### After Optimization f4d27ab04177908254b9b5d058a3ac6f9810a105

| Benchmark                            | Mode | Cnt |   Score |   Error | Units |
| :----------------------------------- | :--- | --: | ------: | ------: | :---- |
| `ParsingBenchmarks.parse_`           | avgt |  25 |  52.451 | ± 0.291 | ns/op |
| `ParsingBenchmarks.parse_a1i`        | avgt |  25 | 164.723 | ± 1.224 | ns/op |
| `ParsingBenchmarks.parse_c`          | avgt |  25 |  52.949 | ± 0.291 | ns/op |
| `ParsingBenchmarks.parse_k1i`        | avgt |  25 | 193.371 | ± 1.875 | ns/op |
| `ParsingBenchmarks.parse_l123456789` | avgt |  25 | 492.890 | ± 5.163 | ns/op |

### DIFF
--- a/src/main/java/com/mojang/brigadier/CommandDispatcher.java
+++ b/src/main/java/com/mojang/brigadier/CommandDispatcher.java
@@ -354,13 +354,15 @@ public class CommandDispatcher<S> {
         Map<CommandNode<S>, CommandSyntaxException> errors = null;
         List<ParseResults<S>> potentials = null;
         final int cursor = originalReader.getCursor();
+        final Collection<? extends CommandNode<S>> relevantNodes = node.getRelevantNodes(originalReader);
+        final boolean unique = relevantNodes.size() == 1;
 
-        for (final CommandNode<S> child : node.getRelevantNodes(originalReader)) {
+        for (final CommandNode<S> child : relevantNodes) {
             if (!child.canUse(source)) {
                 continue;
             }
-            final CommandContextBuilder<S> context = contextSoFar.copy();
-            final StringReader reader = new StringReader(originalReader);
+            final CommandContextBuilder<S> context = unique ? contextSoFar : contextSoFar.copy();
+            final StringReader reader = unique ? originalReader : new StringReader(originalReader);
             try {
                 try {
                     child.parse(reader, context);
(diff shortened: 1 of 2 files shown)


src/main/java/com/mojang/brigadier/CommandDispatcher.java
```diff
@@ -391,16 +393,23 @@ private ParseResults<S> parseNodes(final CommandNode<S> node, final StringReader
                     return new ParseResults<>(context, parse.getReader(), parse.getExceptions());
                 } else {
                     final ParseResults<S> parse = parseNodes(child, reader, context);
+                    if (unique) {
+                        return parse;
+                    }
```
intsuc, line 398: If `unique`, the `potentials` will always contain only `parse` and this `parse` will be returned after exiting this `for` loop, so we can return `parse` here without allocating `potentials`.

src/main/java/com/mojang/brigadier/CommandDispatcher.java
```diff
@@ -391,16 +393,23 @@ private ParseResults<S> parseNodes(final CommandNode<S> node, final StringReader
...
                     if (potentials == null) {
-                        potentials = new ArrayList<>(1);
+                        potentials = new ArrayList<>(relevantNodes.size());
```
intsuc, line 400: Since `potentials` will contain `ParseResults<S>` up to the amount of `relevantNodes.size()`, allocate that capacity in advance to ensure resizing never occurs.

src/test/java/com/mojang/brigadier/benchmarks/ParsingBenchmarks.java
yanisderbikov, line 1: I think this is less readable. But if it increase performance it's ok.

yanisderbikov (approved):
I think this is less readable. But if it increase performance it's ok.

PR #138, opened 2023-08-31
Preserve arguments after redirects

### LINKED ISSUE #137: Arguments in a `CommandContext` disappear when a node redirects
This may well be the intended behavior, but I wanted to check because it feels like a bug. If this is intended, I was wondering what the workaround would be?

---

I encountered this problem when working generally with this sort of command structure:

```
(base command) = A node at the end of some CommandNode structure
(final arguments) = Some CommandNode structure

(intermediate branch) = Some CommandNode structure
(branches) = List of (intermediate branch)

(base command) -> (branches) -> (final arguments)
```

I have some base command path, and I want to add a bunch of branches after that base command node. These branches should converge onto the same final arguments node structure. Abstractly, this is how I achieved that:

```
baseCommand.addChild(firstBranch)

for the rest of the branches:
    branch.redirect(firstBranch)
    baseCommand.addChild(branch)

// Actually build finalArguments
firstBranch.addChild(finalArguments)
```

I setup `(base command) -> (first branch)`, then setup `(base command) -> (branch) -> redirect to (firstBranch)`. Then, I add the final arguments onto the first branch. Since the other branches redirect to first branch, their parsing correctly continues on to the final arguments.

It is important to note that the final arguments structure is built after the branches are built. That's why I redirect the other branches to the first branch. With that, I only need to add the final arguments to the first branch, and therefore I only need to keep track of this "representative" branch in my building code.

I only introduce this as context for the bug. If you think my real problem is that I'm building my command wrong, fair enough. I'm happy to go into more depth about my specific situation if the bug is actually intended behavior.

---

To show this problem, I set up this simple example:

```java
CommandDispatcher<Object> dispatcher = new CommandDispatcher<>();

CommandNode<Object> baseCommand = literal("command").build();

// Add first branch
CommandNode<Object> firstBranchTermination = argument("number", IntegerArgumentType.integer(10, 20)).build();
baseCommand.addChild(literal("high").then(firstBranchTermination).build());

// Redirect other branches to first branch
CommandNode<Object> otherBranch = argument("number", IntegerArgumentType.integer(0, 10)).redirect(firstBranchTermination).build();
baseCommand.addChild(literal("low").then(otherBranch).build());

// Actually build final arguments
CommandNode<Object> finalArgument = argument("string", StringArgumentType.greedyString())
        .executes(context -> {
            int number = context.getArgument("number", Integer.class);
            System.out.println("The given number is " + number);

            String stringArgument = context.getArgument("string", String.class);
            System.out.println("The string was given as " + stringArgument);
            return 1;
        })
        .build();

// Add final arguments at the convergence of the branches
firstBranchTermination.addChild(finalArgument);

// Register command
dispatcher.getRoot().addChild(baseCommand);

// Test cases
dispatcher.execute("command high 11 some text", new Object());
dispatcher.execute("command low 5 some text", new Object());
```

Here's how that connects to my abstract structure before

```
(base command) = literal("command")
(final arguments) = argument("string", GreedyString) with execution defined

(branches):
    (first branch) = literal("high").then(argument("number", Integer from 10 to 20))
    (other branch) = literal("low").then(argument("number", Integer from 0 to 10))
```

There is a low branch that only accepts integers from 0 to 10, and a high branch that only accepts integers from 10 to 20. Both of these branches converge on greedy string, which has an executor that uses the given `"number"` and `"string"` argument.

---

So, the outcome I expect from running this code is that the two test cases at the end execute without failure

```java
dispatcher.execute("command high 11 some text", new Object());
dispatcher.execute("command low 5 some text", new Object());
```

The first case that uses the high branch works as expected. However, the low branch case throws this exception:

``` java
java.lang.IllegalArgumentException: No such argument 'number' exists on this command
	at com.mojang.brigadier.context.CommandContext.getArgument(CommandContext.java:92)
	at com.mojang.brigadier.CustomTest.lambda$test$0(CustomTest.java:34) // This is the `int number = context.getArgument("number", Integer.class);` line
	at com.mojang.brigadier.CommandDispatcher.execute(CommandDispatcher.java:264)
	at com.mojang.brigadier.CommandDispatcher.execute(CommandDispatcher.java:177)
	at com.mojang.brigadier.CommandDispatcher.execute(CommandDispatcher.java:142)
	at com.mojang.brigadier.CustomTest.test(CustomTest.java:49)
```

When I use the redirected `low` branch, the `"number"` argument is not present in the `CommandContext`, hence the title of this issue. Even though the `"number"` argument was parsed as `5`, it is not available when the command executes. When a node redirects, the arguments defined before the redirected node disappear.

---

I think the bug comes from this line in `CommandDispatcher` (The same problem likely occurs when commands fork due to similar code on line 248):

https://github.com/Mojang/brigadier/blob/f20bede62a516a11a468d27d3f1adde2085762bc/src/main/java/com/mojang/brigadier/CommandDispatcher.java#L239

I don't know the reasoning behind the parsing and execution logic here, but I can comment on what I see happening here. I don't know why, but I can tell you how this bug happens.

When `"command low 5 some text"` is parsed, the resulting command context looks something like this:

```
Input: "command low 5 some text"
CommandContext:
    Range 0 to 13 ("command low 5")
    Arguments:
        "number" -> Range 12 to 13 ("5"), result: 5
    Forks: false
    child CommandContext:
        Range 14 to 23 ("some text")
        Arguments:
            "string" -> Range 14 to 23 ("some text"), result: "some text"
        Command: Yes, I have a command to execute
```

The original context parsed everything up to the redirected node. It maps the `"number"` argument to its value, `5`. When parsing reached the redirect, it started a new child context to parse everything after the redirected node. This context maps the `"string"` argument to its value, `"some text"`. The child context also knows that it has a command to execute. Again, I don't know why, but this seems reasonable.

Anyway, the line I linked to above is part of code responsible for using this `CommandContext` to actually execute a command. It looks through all the child contexts and handles all the redirects, especially if the context forks and the executor needs to be multiplied among many sources. The line I specifically linked to is responsible for figuring out the next context in this case with a child context that does not fork. Remember, that line is:

```java
next.add(child.copyFor(context.getSource()));
```

So, the next context that will be evaluated is the child context, but copied to have the original context's source. So, now the context looks like this:

```
CommandContext:
    Range 14 to 23 ("some text")
    Arguments:
        "string" -> Range 14 to 23 ("some text"), result: "some text"
    Command: Yes, I have a command to execute
```

This context ends the chain, and it is executable, so the command is run using the current context. However, the current context only contains the `"string"` argument, not the `"number"` argument. The `"number"` argument only existed in the original context, and it was not copied to the child context. Hence, the `"number"` argument does not exist while running the command, and I get the `IllegalArgumentException`.

---

The main reason why I think this is a bug is because it can be easily fixed. Instead of the current code that just copies the command source:

```java
next.add(child.copyFor(context.getSource()));
```

Using this line solves my issue:

```java
next.add(child.copyFor(context.getSource()).copyArguments(context));
```

Where `CommandContext#copyArguments` is defined like so:

```java
public CommandContext<S> copyArguments(CommandContext<S> context) {
    Map<String, ParsedArgument<S, ?>> newArguments = new HashMap<>();
    newArguments.putAll(context.arguments);
    newArguments.putAll(this.arguments);
    return new CommandContext<>(source, input, newArguments, command, rootNode, nodes, range, child, modifier, forks);
}
```

It seems likely to me that not copying the parent context's arguments is just a semantics error and not intended behavior. 

Due to the lack of documentation for the `redirect` and `fork` methods, I'm not sure what behavior to expect. It makes sense to me that prior arguments would be preserved after a redirect, so I assume this is a bug. Looking at the current uses of redirect and fork, I don't think my changes would cause different behavior. For example, Minecraft's `/execute` command uses redirects to have a looping argument command structure. That means I could reuse the same arguments like so:

```
/execute as @a as @e[type=!player,limit=1] ...
```

Here, the `"targets"` argument is used twice, once as `@a` and once as `@e[type=!player,limit=1]`. I assume the `CommandContext` in this situation would like the following:

```
CommandContext:
    Source: original sender
    Arguments:
        "entities" -> result: @a
    Forks: true
    child CommandContext:
        Arguments:
            "entities" -> result: @e[type=!player,limit=1]
        Forks: true
        child CommandContext:
            ...
```

When the original context's fork is being evaluated, `"entities"` is `@a` as expected. So, the redirect modifier should correctly select all the players as the next sources for the command. When merging the arguments for the new `CommandContext`, the child context's arguments take precedence. So, the next context will look like this:

```
CommandContext:
    Source: All players
    Arguments:
        "entities" -> result: @e[type=!player,limit=1]
    child CommandContext:
        ...
```

Since the child context already had a value for `"entities"`, merging arguments from the parent `CommandContext` doesn't do anything. Therefore, the redirect modifier correctly still processes `@e[type=!player,limit=1]` as the next selector. Overall, merging the arguments doesn't affect old command structures, since if they expect an argument, it will always override the value possibly given to the parent context.

---

So, I hope the fact that arguments are not copied to child contexts is a bug, because I think it would solve my problem. If it is a bug, I think it can be easily fixed without affecting old commands. It seems intuitive to me that arguments would be preserved after redirected nodes, but I can't find any documentation that suggests one way or the other.

If this is a bug, I can definitely make a PR for this. I'm not familiar with the code style of this repo yet, but I think I can figure it out with time. The changes shouldn't be too large; I mostly just need to figure out how to create a test for this.

willkroboth:
This PR resolves #137. 

---

Previously, when executing a `CommandContext` that has a child `CommandContext` due to the parser encountering a redirected node, the arguments present in the original context would not be copied to the child context. In the example I presented in #137, the command structure looked like this:

```java
literal("command"):
    literal("high") -> argument("number", Integer from 10 to 20) (first branch) -> argument("string", greedy string) -> executor
    literal("low") -> argument("number", Integer from 0 to 10) -> redirects to (first branch) node
```

The bug I identified in #137 means that when executing the redirected branch (eg. `"command low 5 some text"`), an `IllegalArgumentException` would be thrown since the executor expects the `"number"` argument to be present in the child `CommandContext`, which is used when executing the command. I expected that the command should run without error, with the value for the `"number"` argument being `5`. In contrast, the 'real' branch would always work as expected, with the `"number"` argument present as given.

---

This PR fixes this bug by adding and using the `CommandContext#copyWithArgumentsOf` method:

```java
public CommandContext<S> copyWithArgumentsOf(CommandContext<S> context) {
    Map<String, ParsedArgument<S, ?>> newArguments = new HashMap<>();
    newArguments.putAll(context.arguments);
    newArguments.putAll(this.arguments);
    return new CommandContext<>(source, input, newArguments, command, rootNode, nodes, range, child, modifier, forks);
}
```

In my example case, the `"number" -> value: 5` argument present in the original `CommandContext` would be copied to the child `CommandContext`, so that it is present when the executor is run.

---

I added tests that directly make sure the new `CommandContext#copyWithArgumentsOf` method works as expected. I want to add tests to make sure `copyWithArgumentsOf` is being used correctly when executing commands, but I wasn't sure where to put a test like that. I wanted to make sure that I followed the code style of this repo. I'm not yet sure how the code base tests functionality like this, so I erred on the side of not adding those tests yet.

---

If you'd like to discuss whether or not the 'bug' this PR resolves is actually intended behavior, I think those comments fit best as replies to the original issue, #137. I'm honestly not sure if my issue is intended behavior or not since I couldn't find any documentation on what node redirects are supposed to do. It seems most intuitive to me that the arguments should be preserved, hence why I went through the effort to make this PR :P.

### DIFF
--- a/src/main/java/com/mojang/brigadier/CommandDispatcher.java
+++ b/src/main/java/com/mojang/brigadier/CommandDispatcher.java
@@ -236,7 +236,7 @@ public class CommandDispatcher<S> {
                             if (next == null) {
                                 next = new ArrayList<>(1);
                             }
-                            next.add(child.copyFor(context.getSource()));
+                            next.add(child.copyFor(context.getSource()).copyWithArgumentsOf(context));
                         } else {
                             try {
                                 final Collection<S> results = modifier.apply(context);
@@ -245,7 +245,7 @@ public class CommandDispatcher<S> {
                                         next = new ArrayList<>(results.size());
                                     }
                                     for (final S source : results) {
-                                        next.add(child.copyFor(source));
+                                        next.add(child.copyFor(source).copyWithArgumentsOf(context));
                                     }
                                 } else {
                                     foundCommand = true;
--- a/src/main/java/com/mojang/brigadier/context/CommandContext.java
+++ b/src/main/java/com/mojang/brigadier/context/CommandContext.java
@@ -57,6 +57,13 @@ public class CommandContext<S> {
         return new CommandContext<>(source, input, arguments, command, rootNode, nodes, range, child, modifier, forks);
     }
 
+    public CommandContext<S> copyWithArgumentsOf(CommandContext<S> context) {
+        Map<String, ParsedArgument<S, ?>> newArguments = new HashMap<>();
+        newArguments.putAll(context.arguments);
+        newArguments.putAll(this.arguments);
+        return new CommandContext<>(source, input, newArguments, command, rootNode, nodes, range, child, modifier, forks);
+    }
+
     public CommandContext<S> getChild() {
         return child;
     }
--- a/src/test/java/com/mojang/brigadier/context/CommandContextTest.java
+++ b/src/test/java/com/mojang/brigadier/context/CommandContextTest.java
@@ -52,6 +52,34 @@ public class CommandContextTest {
     }
 
     @Test
+    public void testCopyArgumentsFills() throws Exception {
+        final CommandContext<Object> originalContext = new CommandContextBuilder<>(dispatcher, source, rootNode, 0).withArgument("foo", new ParsedArgument<>(0, 1, 123)).build("123");
+        final CommandContext<Object> childContext = new CommandContextBuilder<>(dispatcher, source, rootNode, 0).build("123");
+
+        final CommandContext<Object> copiedContext = childContext.copyWithArgumentsOf(originalContext);
+        assertThat(copiedContext.getArgument("foo", int.class), is(123));
+    }
+
+    @Test
+    public void testCopyArgumentsOverrides() throws Exception {
+        final CommandContext<Object> originalContext = new CommandContextBuilder<>(dispatcher, source, rootNode, 0).withArgument("foo", new ParsedArgument<>(0, 1, 123)).build("123");
+        final CommandContext<Object> childContext = new CommandContextBuilder<>(dispatcher, source, rootNode, 0).withArgument("foo", new ParsedArgument<>(0, 1, "123")).build("123");
+
+        final CommandContext<Object> copiedContext = childContext.copyWithArgumentsOf(originalContext);
+        assertThat(copiedContext.getArgument("foo", String.class), is("123"));
+    }
+
+    @Test
+    public void testCopyArgumentsMerges() throws Exception {
+        final CommandContext<Object> originalContext = new CommandContextBuilder<>(dispatcher, source, rootNode, 0).withArgument("foo", new ParsedArgument<>(0, 1, 123)).build("123");
+        final CommandContext<Object> childContext = new CommandContextBuilder<>(dispatcher, source, rootNode, 0).withArgument("bar", new ParsedArgument<>(0, 1, "123")).build("123");
+
+        final CommandContext<Object> copiedContext = childContext.copyWithArgumentsOf(originalContext);
+        assertThat(copiedContext.getArgument("foo", int.class), is(123));
+        assertThat(copiedContext.getArgument("bar", String.class), is("123"));
+    }
+
+    @Test
     public void testSource() throws Exception {
         assertThat(builder.build("").getSource(), is(source));
     }


yanisderbikov:
If you will add a summary it will be better to understand your main message :)

willkroboth:
Sure. I think I also found a different but related problem with redirects, suggestions, and arguments, so I'll explain that here too.

---

### Summary: 

When _suggesting_ arguments after a redirect, the `CommandContext` given can only access arguments _before_ the redirect.

When _executing_ a command after a redirect, the `CommandContext` given can only access arguments _after_ the redirect.

These two situations handle arguments differently when redirects are present. I believe that both of these situations are bugs. The correct outcome should be that all arguments are present.

---

### Reproducing the 'bugs'

To help explain this better, I created a demonstration here https://github.com/willkroboth/brigadier/commit/b1663e517b867c454bb8f74651986b9048b982bc that looks like this:

```java
@Test
public void testCustomStuff() throws Exception {
    System.out.println("Testing custom stuff");

    SuggestionProvider<Object> monitorSuggestions = (context, builder) -> {
        System.out.println(context.getArguments());

        builder.suggest("a").suggest("b").suggest("c");
        return builder.buildFuture();
    };

    ArgumentCommandNode<Object, String> end = argument("ending", StringArgumentType.word())
            .suggests(monitorSuggestions)
            .then(argument("lastArg", StringArgumentType.word())
                    .suggests(monitorSuggestions)
                    .executes(context -> {
                        System.out.println(context.getArguments());

                        return 1;
                    })
            )
            .build();

    ArgumentCommandNode<Object, Integer> firstBranchEnding = argument("number", IntegerArgumentType.integer(1, 10)).then(end).build();
    LiteralArgumentBuilder<Object> firstBranch = literal("low").then(firstBranchEnding);

    LiteralArgumentBuilder<Object> secondBranch = literal("high").then(
            argument("number", IntegerArgumentType.integer(11, 20))
                    .redirect(firstBranchEnding)
    );

    LiteralArgumentBuilder<Object> base = literal("base")
            .then(firstBranch)
            .then(secondBranch);

    subject.register(base);

    System.out.println(subject.getCompletionSuggestions(subject.parse("base low 5 ", source)).get());
    System.out.println();
    System.out.println(subject.getCompletionSuggestions(subject.parse("base low 5 a ", source)).get());
    System.out.println();
    System.out.println(subject.getCompletionSuggestions(subject.parse("base high 15 ", source)).get());
    System.out.println();
    System.out.println(subject.getCompletionSuggestions(subject.parse("base high 15 a ", source)).get());
    System.out.println();

    subject.execute("base low 5 a b", source);
    System.out.println();
    subject.execute("base high 15 a b", source);
    System.out.println();
}
```

When this test is run, this gets printed into the console:
```java
Testing custom stuff
// First test
{number=ParsedArgument{range=StringRange{start=9, end=10}, result=5}}
Suggestions{range=StringRange{start=11, end=11}, suggestions=[Suggestion{range=StringRange{start=11, end=11}, text='a', tooltip='null'}, Suggestion{range=StringRange{start=11, end=11}, text='b', tooltip='null'}, Suggestion{range=StringRange{start=11, end=11}, text='c', tooltip='null'}]}

{number=ParsedArgument{range=StringRange{start=9, end=10}, result=5}, ending=ParsedArgument{range=StringRange{start=11, end=12}, result=a}}
Suggestions{range=StringRange{start=13, end=13}, suggestions=[Suggestion{range=StringRange{start=13, end=13}, text='a', tooltip='null'}, Suggestion{range=StringRange{start=13, end=13}, text='b', tooltip='null'}, Suggestion{range=StringRange{start=13, end=13}, text='c', tooltip='null'}]}

// Second test
{number=ParsedArgument{range=StringRange{start=10, end=12}, result=15}}
Suggestions{range=StringRange{start=13, end=13}, suggestions=[Suggestion{range=StringRange{start=13, end=13}, text='a', tooltip='null'}, Suggestion{range=StringRange{start=13, end=13}, text='b', tooltip='null'}, Suggestion{range=StringRange{start=13, end=13}, text='c', tooltip='null'}]}

{number=ParsedArgument{range=StringRange{start=10, end=12}, result=15}}
Suggestions{range=StringRange{start=15, end=15}, suggestions=[Suggestion{range=StringRange{start=15, end=15}, text='a', tooltip='null'}, Suggestion{range=StringRange{start=15, end=15}, text='b', tooltip='null'}, Suggestion{range=StringRange{start=15, end=15}, text='c', tooltip='null'}]}

// Third test
{number=ParsedArgument{range=StringRange{start=9, end=10}, result=5}, ending=ParsedArgument{range=StringRange{start=11, end=12}, result=a}, lastArg=ParsedArgument{range=StringRange{start=13, end=14}, result=b}}

// Fourth test
{ending=ParsedArgument{range=StringRange{start=13, end=14}, result=a}, lastArg=ParsedArgument{range=StringRange{start=15, end=16}, result=b}}
```

This output shows both 'bugs' in action. It probably needs an explanation though if you don't know what to look for.

---

### Structure of the example

The structure of my test command looks like this:

```basic
literal base
-> then literal low
   -> then argument number (integer from 0 to 10)
      -> then argument ending (string that's just a single word)
         suggests 'monitorSuggestions'
         -> then argument lastArg (string that's just a single word)
            suggests 'monitorSuggestions'
            executes the only executor for this command
 -> then literal high
    -> then argument number (integer from 11 to 20)
       redirects to base->low->number
```

`monitorSuggestions` looks like this:

```java
SuggestionProvider<Object> monitorSuggestions = (context, builder) -> {
    System.out.println(context.getArguments());

    builder.suggest("a").suggest("b").suggest("c");
    return builder.buildFuture();
};
```

And the `Command` on `lastArg` looks like this:

```
context -> {
    System.out.println(context.getArguments());

    return 1;
}
```

---

### The problem with redirects, arguments, and _suggestions_

The first test checks the suggestions for `"base low 5 "` and `"base low 5 a "`. Since this path does not include a redirect, it acts as expected. The `CommandContext` when building suggestions includes both the `number` and `ending` arguments.

Simplified output:
```
// Suggesting `"base low 5 "`
Arguments: {number=5}
Suggestions{'a', 'b', 'c'}

// Suggesting `"base low 5 a"`
Arguments: {number=5, ending=a}
Suggestions{'a', 'b', 'c'}
```

The second test checks the suggestions for `"base high 15 "` and `"base high 15 a "`. This path includes a redirect, so the arguments get messed up.

Simplified output:
```
// Suggesting `"base high 15 "`
Arguments: {number=15}
Suggestions{'a', 'b', 'c'}

// Suggesting `"base high 15 a"`
Arguments: {number=15}
Suggestions{'a', 'b', 'c'}
```

Here, the `CommandContext` used when building the suggestions for `"base high 15 a"` is missing the value for the argument `ending`, even though it was given as `"a"`. When `base->high->number` redirects to `base->low->number->ending`, only arguments **before** the redirect are accessible.

---

### The problem with redirects, arguments, and _executes_

The third test checks what happens when executing `"base low 5 a b"`. Since this path does not include a redirect, it acts as expected. The `CommandContext` includes the `number`, `ending` and `lastArg` arguments.

Simplified output:
```
Arguments: {number=5, ending=a, lastArg=b}
```

The fourth test checks what happens when executing `base high 15 a b`. This path includes a redirect, so the arguments get messed up.

Simplified output:
```
Arguments: {ending=a, lastArg=b}
```

Here, the `CommandContext` used when executing `"base high 15 a b"` is missing the value for the argument `number`, even though it was given as `15`. When `base->high->number` redirects to `base->low->number->ending`, only arguments **after** the redirect are accessible.

---

I think that both of these behaviors are unintentional. It seems that arguments are not being handled properly when redirects (and forks) create child `CommandContext` objects. This PR aims to resolve these two 'bugs' by making sure that all arguments given in a command are present when redirects are involved, the same way it works with a linear path when there are no redirects.

It looks like https://github.com/Mojang/brigadier/pull/140 moved some functionality around, so this PR currently conflicts with the main branch. I also recently realized there was a bug with the suggestions (thanks to https://github.com/JorelAli/CommandAPI/issues/310), so I haven't figured out how to fix that yet. I'll try to figure something out though and update this PR soon.

PR #145, opened 2024-01-05
Allow custom predicate that replaces `StringReader#isAllowedInUnquotedString`

astro-angelfish:
Instead of https://github.com/Mojang/brigadier/pull/131 with greedy charset solution, I tried another approach that allows users to define their own whitelist. It might be useful when users trying to define their own syntax on command arguments like entity selectors or something else that might accept non-ASCII characters

Usage of `StringReader#readString` or `StringReader#readUnquotedString` without any argument would lead to the default behavior, which is calling `StringReader#isAllowedInUnquotedString` as their whitelist predicate.

Fixes https://github.com/Mojang/brigadier/issues/103.

### DIFF
--- a/src/main/java/com/mojang/brigadier/StringReader.java
+++ b/src/main/java/com/mojang/brigadier/StringReader.java
@@ -5,6 +5,8 @@ package com.mojang.brigadier;
 
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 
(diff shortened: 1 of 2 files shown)


Privatech38:
This will cause problems when dealing with larger whitelists/list of allowed chars, therefor consider merging yours and #131's way of dealing with this issue.

PR #153, opened 2024-08-15
Relax string() type and add relaxedWord()

### LINKED ISSUE #152: More generic string argument type
At the moment there are three variants for string argument types. But two of them only allow a very limited set of characters, and greedy string is not really applicable.

A more generic string argument type would be welcome which doesn't have those limitations. This would be useful especially for names which don't represent files, etc.

valaphee:
Closes https://github.com/Mojang/brigadier/issues/152
by modifying the `string()` type to be more relaxed as it already allows all characters if escape, and this just makes it more convenient to use by not requiring quoting except for white spaces, escapes and quotes.

Also adds a new string type called `relaxedWord()`. Which behaves the same as string except not allowing quoting and therefore also white spaces. `word()` is kept unchanged as this would change the possible inputs and may break things.

### DIFF
--- a/src/main/java/com/mojang/brigadier/StringReader.java
+++ b/src/main/java/com/mojang/brigadier/StringReader.java
@@ -174,9 +174,21 @@ public class StringReader implements ImmutableStringReader {
             || c == '.' || c == '+';
     }
 
+    public static boolean isAllowedInUnquotedStringRelaxed(final char c) {
+        return c != ' ' && c != '\\' && c != '"';
+    }
+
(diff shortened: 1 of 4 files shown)


valaphee:
The CLA bot doesn't seem to work?

StillLutto (approved):
This looks good to me, please get this approved, this is probably single-handedly the most useful argument for plugin developers.

Clexus:
Really want to see this get merged, but brigadier is a dead project now

PR #155, opened 2024-11-08
Add missing `fail`s to end of `try` blocks in tests

bbb651:
I'm working on changes to `read{Int,Long,Float,Double}` to make them less greedy and noitced some tests that should've failed.

I went through every try catch and verified `try` properly `fail`s.

### DIFF
--- a/src/test/java/com/mojang/brigadier/StringReaderTest.java
+++ b/src/test/java/com/mojang/brigadier/StringReaderTest.java
@@ -240,6 +240,7 @@ public class StringReaderTest {
     public void readQuotedString_noOpen() throws Exception {
         try {
             new StringReader("hello world\"").readQuotedString();
+            fail();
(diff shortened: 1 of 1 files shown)


bbb651:
I finished implementing my [primitive parsing rework](https://github.com/bbb651/brigadier/tree/accurate-primitives), making a PR is blocked on this. Is it ok to remove `isAllowedNumber`? There are [at least a few usages of it, mainly in minecraft mods](https://github.com/search?q=language%3Ajava+StringReader.isAllowedNumber%28&type=code), I think it's quite unfortunate it was public in the first place (on that note, should `isAsciiDigit` be private?)

By the way, I noticed `readBoolean` uses `readString()` rather than `readUnquotedString()` directly, making it parse `"true"` and `"false"` in addition to `true` and `false`. Is this intended? It's untested and undocumented.

PR #156, opened 2024-12-08
Fix incorrect context being used for suggestions (MC-278771)

senseiwells:
This fixes [MC-278771](https://bugs.mojang.com/browse/MC-278771).

The overview of this issue is that `CommandDispatcher#getCompletionSuggestions` doesn't provide the correct `CommandContext` to the `CommandNode` for which it's listing suggestions from. This is an issue as some suggestions may be contextual and rely on previous arguments in the command tree. Currently the root `CommandContext` is provided, this means that any redirect causes the incorrect context to be provided and thus an `IllegalStateException` to be thrown as the argument cannot be found.

The solution to this is to provide the correct `CommandContextBuilder` in the `CommandContextBuilder#findSuggestionContext` method. This builder can then provide the correct `CommandContext` for the respective command node.


Gegy (approved):
Looks great, thank you!!

Dattish (approved):

PR #157, opened 2025-03-24
Don't suggest nodes which the source does not have requirements for

willkroboth:
Right now, requirements are not checked when generating suggestions. This means that the following test fails:

```java
// Added to `CommandSuggestionsTest`
@Test
public void getCompletionSuggestions_unmetRequirement() throws Exception {
    subject.register(
        literal("parent")
            .then(
                literal("invalid").requires(o -> false)
            )
            .then(
                literal("valid")
            )
    );

    final ParseResults<Object> parse = subject.parse("parent ", source);
    final Suggestions result = subject.getCompletionSuggestions(parse).join();

    // java.lang.AssertionError: 
    //  Expected: <[Suggestion{range=StringRange{start=7, end=7}, text='valid', tooltip='null'}]>
    //  but: was <[Suggestion{range=StringRange{start=7, end=7}, text='invalid', tooltip='null'}, 
    //  Suggestion{range=StringRange{start=7, end=7}, text='valid', tooltip='null'}]>
    assertThat(result.getList(), equalTo(Lists.newArrayList(new Suggestion(StringRange.at(7), "valid"))));
}
```

Since the `invalid` literal was given the requirement `o -> false`, the source cannot execute that node or any of its children. However, `invalid` is still given as a suggestion. Since the source is not allowed to use this node, it should not have it suggested, since that confusingly suggests it could work.

This PR fixes this issue by ignoring a node's suggestions if the source does not meet its requirements.

### DIFF
--- a/src/main/java/com/mojang/brigadier/CommandDispatcher.java
+++ b/src/main/java/com/mojang/brigadier/CommandDispatcher.java
@@ -525,6 +525,7 @@ public class CommandDispatcher<S> {
 
     public CompletableFuture<Suggestions> getCompletionSuggestions(final ParseResults<S> parse, int cursor) {
         final CommandContextBuilder<S> context = parse.getContext();
+        S source = context.getSource();
 
         final SuggestionContext<S> nodeBeforeCursor = context.findSuggestionContext(cursor);
         final CommandNode<S> parent = nodeBeforeCursor.parent;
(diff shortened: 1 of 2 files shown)


sakurawald:
Look good to me. This pull-request works perfect.

Currently, the results of `/help <cmd>` and `client-side tab completion` are `in-consistent`.
The `/help` command calls the `ComamndNode#canUse`, while the `required argument with custom suggestions provider` will not call the `CommandNode#canUse`.

joelwaelchli (approved):

sakurawald:
I am happy to see there is progress in this pull-request.

Lioneldellarossa (approved):

